PR #12, opened 2020-06-02
Update Cocoapods instructions

nickromano:


ios/README.md
```diff
@@ -1,14 +1,12 @@
 # FluentIcons iOS
 
-![fluent design icons](../../art/readme-asset.png)
```
nickromano, line 3: This wasn't loading correctly but I figured I'll just remove it for now since we already have it on the main readme.

PR #43, opened 2020-07-03
Add python script to generate BUILD.gn used by chromium Android project

huilliu:
Chromium projects use GN as Makefile to build the whole project. Add BUILD.gn for fluent icon, so that we can integrate it  to Edge Android project conveniently. There is a GN target named "fluent_icon_resources" in the BUILD.gn. If other module need to use the fluent icon resources, we only need add a dependence to "fluent_icon_resources" in other BUILD.gn file.


willhou:
@huilliu can you exclude the `android/BUILD.gn` file from this change as it will be generated and committed during the CI process.

PR #55, opened 2020-07-21
Nuget ado pipeline

isatoderici:
[Package Management] Create and publish NuGet package:

- Create .nuspec file

- Generate build number

- Select current version of Xcode

- Prepare NuGet job for ADO pipeline:
  1. Configure Xcode build
  2. Add build number
  3. Select Xcode version
  4. Configure NuGet to publish job

- Prepare NuGet package and customize the archived build


scripts/prepare_for_nuget_pack.sh
```diff
@@ -0,0 +1,67 @@
...
+# Rename the FluentIcons_Example-Swift.h to FluentIcons-Swift.h for consistency with Framework consumption
+echo "Copy iOS generated Swift header to include dir"
+rsync -a "DerivedData/Build/Intermediates.noindex/FluentIcons.build/Release-iphonesimulator/FluentIcons_Example.build/DerivedSources/FluentIcons_Example-Swift.h" "$NUGET_OUTPUT_INCLUDE_DIR_IOS_FLUENTICONS/FluentIcons-Swift.h"
+
+# cd into the products directory to make copying all the output easier
+cd $PRODUCTS_DIR
+
+make_dir_if_necessary "nuget/Debug-iphoneos"
+echo "Copy Debug-iphoneos build output into nuget folder"
+rsync -a Debug-iphoneos/FluentIcons_Example.swiftmodule/ nuget/Debug-iphoneos/FluentIcons_Example.swiftmodule/
```
markavitale, line 46: A couple issues:

- I think we're copying the example consumer framework, not the framework itself. Kind of like copying a test app instead of the library we want.
- Looks like we're only copying the swiftmodule. We'll need to copy the library itself (.a) or a .framework if that's what we're copying right now. A swiftmodule is kind of the swift equivalent of headers, and doesn't actually contain the built binary we care about.

DerivedData/Build/Products/Debug-iphonesimulator/FluentIcons-iOS/FluentIcons.framework/_CodeSignature/CodeResources
```diff
@@ -0,0 +1,252 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
markavitale, line 1: Looks like some build output has snuck in here. Can we get rid of that and update the gitignore file?

scripts/FilterIconsCatalog.py
```diff
@@ -0,0 +1,185 @@
+import os
+import shutil
+import zipfile
+
+set = {
```
anandrajeswaran, line 5: The filter list of icons we (one consumer) happen to care about probably shouldn't happen in the repo itself. We should discuss offline if it would be better to clone and add filtering in our clone

scripts/xcode_select_current_version.sh
```diff
@@ -0,0 +1,3 @@
+# Select the current officially supported version of Xcode with the naming convention used by GitHub Actions/Azure DevOps Build Pipelines
```
anandrajeswaran, line 1: If we're just copying images do we need Xcode at all?

isatoderici: Nope, forgot to remove it.

DerivedData/Build/Products/Debug-iphonesimulator/FluentIcons-iOS/FluentIcons.framework/_CodeSignature/CodeResources
isatoderici: Sure, don't know how this file ended up here. :)

JDOG30: #140

anandrajeswaran:
@isatoderici can we close this PR?

isatoderici:
> @isatoderici can we close this PR?

Yes.

PR #63, opened 2020-07-30
Show all weights in icon sheet

nickromano:
Currently, we group all icons by their name, this change now groups them by weight so you can visualize all of the icons.

**Before**
Screen Shot 2020-07-29 at 5 58 59 PM

**After**
Screen Shot 2020-07-29 at 5 59 43 PM


willhou:
👍 ✨

PR #94, opened 2020-10-02
Jacust/2020 10 02 bold new update

jasoncuster:
Added variants to Bold for locale specific icons.


willhou:
@jasoncuster For languages of the same locale but different region, e.g. "fr-ca" and "fr-fr", can we replace them with one single folder "fr" since the icons are the same? This will help us simplify and reduce the library size. Other ones are "es-es" and "es-mx", "sr-cyrl-cs" and "sr-cyrl-rs" and "sr-latn-cs" and "sr-latn-rs".

spencer-nelson:
There are a lot and we were on the fence about which would be preferable. The default behavior for iOS is that separate locales get their own folder, regardless, and we didn't want to force the platform code/end product teams to handle all that, especially if only certain locales are supported where others aren't by product teams. Also, while you can look at it right now and say it's the same for one particular icon we can't guarantee that will be the case for every language/locale variant with every icon.
The simplest route was to have all locales separate, so that's where we ended. If you are confident that won't come back to bite us, we can easily consolidate any/all dupes. If we take that route, how should we change the underlying folder structure?

willhou:
@spencer-nelson On iOS. we could define folder specified by just the language code and then we should only drill down to use a folder with the regional code when the icons are different per regional locale. This is identical to how resource buckets work, in a cascading manner rather than duplicating for each region.

Take French for example, since the icons are the same, they should be placed in the folder with the "fr" locale code, and in the future if there are icons that are different between French French and Canadian French or Swiss French, then we can place that new icon into regional specific folders, "fr_FR", "fr_CA", "fr_CH".

Supported language and locale codes
https://github.com/apple/swift-corelibs-foundation/blob/main/CoreFoundation/PlugIn.subproj/CFBundle_Locale.c

spencer-nelson (approved):

PR #96, opened 2020-10-07
iOS: Remove all cases option to improve compile time

nickromano:
We previously used `allCases` to power the demo mac app, now that we've replaced it with a [generated icon page](https://github.com/microsoft/fluentui-system-icons/blob/master/icons.md) we no longer need the functionality.  Removing it also saves a lot of compile time.

f41bee6e-b01e-4926-bc1a-8cb6349c08e0

### DIFF
--- a/importer/process_ios_assets.py
+++ b/importer/process_ios_assets.py
@@ -216,15 +216,6 @@ def process_assets():
                     )
                 )
 
-        all_cases = []
-        for name, list_of_icons in sorted(all_icons_by_name.items()):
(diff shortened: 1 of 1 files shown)


spencer-nelson (approved):
I assume this isn't used by anyone else, but let's watch the Issues for this being breaking for anyone.

nickromano:
Definitely 👍 thanks @spencer-nelson.  I was thinking I could separate it into another file and use a Pod subspec but didn't want to go through the effort until I knew others were relying on it.

PR #121, opened 2020-11-06
Fix inccorect version mapping on publish.yml

aakash1313:
Testing first on the pr.yml and then will make the change on publish.yml

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -102,7 +102,7 @@ jobs:
 
     - name: Replace version number in flutter/pubspec.yaml
       run: |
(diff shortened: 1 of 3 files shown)


.github/workflows/pr.yml
```diff
@@ -100,6 +100,11 @@ jobs:
       run: npm run deploy:flutter
       working-directory: importer
 
+    - name: Replace version number in flutter/pubspec.yaml
+      run: |
+        sed -i.bk -r "version: s/[0-9]+\.[0-9]+\.[0-9]+/$NEW_VERSION/g" flutter/pubspec.yaml
```
jamesblasco, line 105: I think it should be `sed -i.bk -r "version: s/[0-9]+\.[0-9]+\.[0-9]+/version: $NEW_VERSION/g" flutter/pubspec.yaml`

willhou: #122

PR #122, opened 2020-11-06
Update regex to include version inside the replacement flag

aakash1313:


willhou (approved):

jamesblasco:
@aakash1313 @willhou notice the `version:`  word was removed in the file in previous merge, it needs to be added again in pubspec.yaml

willhou:
@jamesblasco yes you are right, I created #123

PR #123, opened 2020-11-06
Fix version in pubspec.yaml

willhou:
* Add back "version: " string removed by previous script errors #121 and #122
* Manually up version to 1.1.80


jamesblasco:
Perfect! Thank you 🙂

PR #124, opened 2020-11-06
Automating the publish of flutter package on pub.dev

aakash1313:
The purpose of this PR is to automatically publish the fluent ui package on pub.dev whenever a new version is available. This will solve the problem where our flutter users weren't getting the latest icons. With this PR in , they will have always have the latest package.

We perform a dry-run whenever a PR is pushed which is basically creating the package and make sure nothing breaks on the branch. The actual publishing happens when the PR is merged into master.


.github/workflows/pr.yml
```diff
@@ -99,7 +99,3 @@ jobs:
     - name: Run generate script
       run: npm run deploy:flutter
       working-directory: importer
-
-    # TODO: https://github.com/microsoft/fluentui-system-icons/issues/34
-    # - name: Build Flutter library
-    #   uses: ?
```
willhou, line 105: We should run the `flutter build` command here to ensure the library builds without failure?

flutter/CHANGELOG.md
```diff
@@ -1 +1 @@
-## [1.1.82] - Complete change log present here https://github.com/microsoft/fluentui-system-icons/tags
\ No newline at end of file
+## [1.1.83] - Complete change log present here https://github.com/microsoft/fluentui-system-icons/tags
```
willhou, line 1: This should remain as 82? Because the publish script will auto increment and commit it.

flutter/pubspec.yaml
```diff
@@ -1,6 +1,6 @@
 name: fluentui_system_icons
 description: Fluent UI System Icons are a collection of familiar, friendly and modern icons from Microsoft.
-version: 1.1.82
+version: 1.1.83
```
willhou, line 3: This should remain as 82? Because the publish script will auto increment and commit it.

.github/workflows/publish.yml
```diff
@@ -67,9 +67,14 @@ jobs:
...
       working-directory: importer
 
-    # TODO: https://github.com/microsoft/fluentui-system-icons/issues/34
-    # - name: Build and publish Flutter library
-    #   uses: ?
+    # The name should be same as the package name on pub.dev
+    - name: 'fluentui_system_icons'
+      uses: k-paxian/dart-package-publisher@master
+      with:
+        relativePath: '././flutter'
```
willhou, line 74: Do we need `././` or can this be just `flutter`?

willhou:
Can you fill out the PR description? Thanks.

flutter/CHANGELOG.md
aakash1313: @willhou : I used that for mock testing if the auto publish was successful, can we publish with 84 ?

.github/workflows/pr.yml
aakash1313: Added the dry run script to make sure that package is built properly but not publish. Let me know what you think about that :)

.github/workflows/publish.yml
aakash1313: Updated. :)

flutter/CHANGELOG.md
willhou: No we cannot custom a version here, this needs to match what the CI's run count is, so that it stays in sync with the Android and iOS libraries versions.

.github/workflows/pr.yml
willhou: ok, but for dry run, does it need the accessToken and refreshToken? Could we remove them if possible?

.github/workflows/pr.yml
```diff
@@ -100,6 +100,12 @@ jobs:
       run: npm run deploy:flutter
       working-directory: importer
 
-    # TODO: https://github.com/microsoft/fluentui-system-icons/issues/34
-    # - name: Build Flutter library
-    #   uses: ?
\ No newline at end of file
+      # The name should be same as the package name on pub.dev
```
willhou, line 103: Can you add `# Build Flutter library` above this line?

flutter/CHANGELOG.md
aakash1313: I can change it back to 82, but since pub.dev already has 83, (since I did that manually to test if it doesn't fail there), we might have to bump the version 84, so it publishes the changes there.

.github/workflows/pr.yml
aakash1313: It seems accessToken and refreshToken are required for the dry-run . Just tested without that and it failed.

flutter/CHANGELOG.md
willhou: Can you remove 83 from pub.dev? We should have the CI publish 83 to pub.dev

willhou: Is it possible to do that by using this downgrade command? https://dart.dev/tools/pub/cmd/pub-downgrade

aakash1313: @willhou : It is not possible to downgrade the package once it is published. The given link is used to downgrade the dependencies used inside a particular package :/

willhou (approved):

PR #140, opened 2020-12-03
Remove conformance to Equatable and CaseIterable to improve compile time

nickromano:
Not many apps are using these and they affect compile time. If an app does want to use this they can add the following in their app instead:

```swift
extension FluentIcon: Equatable, CaseIterable {}
```


spencer-nelson (changes requested):
Can you please add information on the Equatable and CaseIterable extension in the iOS Readme so the information is preserved?

nickromano:
@spencer-nelson, these were leftover from the demo app, they aren't used for the library. We realized recently that it has a large overhead for compile-time so that was why I was looking to clean it up.

spencer-nelson (approved):

PR #160, opened 2021-01-14
Add publish setup for plain svg icons in NPM

bheston:
Adds a new package with configuration to publish optimized plain svgs using NPM.
- Added system-icons-svg package, named based on discussion about maintaining association with this library and multiple output formats
- Added steps to pr and publish workflows
- Adjusted `generate` script to separate Android specific assets


bheston:
I've not setup `npm publish` or workflow scripts like this before. I've verified the build scripts produce the desired output in `/dist` but please carefully review the workflow and publishing configuration.

bheston:
The intent of publishing this format is to be able to use `svg-inline-loader` in webpack config and reference an icon in code like
```ts
import AddIcon from "@fluentui/system-icons-svg/dist/ic_fluent_add_20_filled.svg";
```

Do underscores seem right, or is there any reason to switch to hyphen for this format?

spencer-nelson:
> The intent of publishing this format is to be able to use `svg-inline-loader` in webpack config and reference an icon in code like
> 
> ```ts
> import AddIcon from "@fluentui/system-icons-svg/dist/ic_fluent_add_20_filled.svg";
> ```
> 
> Do underscores seem right, or is there any reason to switch to hyphen for this format?

The Android and iOS libraries use underscores for the filenames and it's what the raw files use, so I suggest underscores here for consistency.

spencer-nelson (changes requested):
This largely looks good. Please rename the svg set as @fluentui/svg-icons.

bheston:
To further cleanup the import line I removed "ic_fluent_" from the file names. The information is already in the package name and this helps by shortening the amount to type before auto-complete listing.

spencer-nelson (approved):

PR #170, opened 2021-01-29
Fix the npm publish step

spencer-nelson:


.github/workflows/publish.yml
```diff
@@ -126,13 +126,15 @@ jobs:
...
 
     ## Publish svg icons
-    - name: Run svg publish
-      run: npm publish
+    - name: Build and publish SVGs to npm
+      run: |
+        npm install
+        npm config set //registry.npmjs.org:_authToken ${NODE_AUTH_TOKEN}
+        npm run build
+        npm publish --ignore-scripts
```
bheston, line 134: I suspected the order might be off, like `with` needed to be before `run`. Curious here why the `ignore-scripts` was necessary. Seems like the `prepublishOnly` script is more encapsulated.

jasoncuster (approved):

willhou (commented):
LGTM

PR #182, opened 2021-02-13
React library update

spencer-nelson:
Process the components through svgr and tidy up the export with props you can pass to style the fill color.


spencer-nelson:
The resulting components will look like:

import * as React from "react";
interface IAccessTime24Regular {
  primary?: string;
}

const AccessTime24Regular = (props: IAccessTime24Regular) => {
  return <svg width={24} height={24} fill="none" xmlns="http://www.w3.org/2000/svg"><path d="<snip/>" fill={props.primary} /><path d="<snip>" fill={props.primary} /></svg>;
};

export { IAccessTime24Regular, AccessTime24Regular };

packages/react-icons/template.js
```diff
@@ -0,0 +1,29 @@
...
+	componentName.name = componentName.name.substring(3)
+	interfaceName = 'I' + componentName.name
+	combinedExport = '{ ' + interfaceName + ', ' + componentName.name + ' }'
+	exports.declaration.name = componentName.name
+
+	return tpl.ast`
+		${imports}
+
+		interface ${interfaceName} {
+			primary?: string
```
zannstpierre, line 18: You might want to expand this to accept more interface properties.

spencer-nelson: Which properties could/should be needed and why? The only possible ones I can think of would be width/height but those are inherent in the selected icon.

packages/react-icons/template.js
```diff
@@ -0,0 +1,29 @@
...
+	
+	componentName.name = componentName.name.substring(3)
+	interfaceName = 'I' + componentName.name
+	combinedExport = '{ ' + interfaceName + ', ' + componentName.name + ' }'
+	exports.declaration.name = componentName.name
+
+	return tpl.ast`
+		${imports}
+
+		interface ${interfaceName} {
```
spencer-nelson, line 17: Rather than a different interface for each component, would it be better to have one single IFluentIconProps interface that is just used by all of them? That seems more inline with the idea that all the icons are part of a cohesive system.

zannstpierre: If they could use all use it, that would be ideal.

packages/react-icons/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "spnels-fluent-system-test0000",
+  "version": "1.1.50",
+  "description": "Fluent System Icons are a collection of familiar, friendly, and modern icons from Microsoft.",
+  "author": "Microsoft Inc.",
```
Philip-Scott, line 5: Do you need this author tag? I see that it's been removed on a separate PR here: https://github.com/microsoft/fluentui-system-icons/pull/177

packages/react-icons/template.js
Philip-Scott: I agree with this comment. If all the props are the same for all icons, a single shared IFluentIconProps interface would be ideal and allow for a simpler type usage downstream. Then later on if an icon has it's own props, you could have a union type, kind of like `IFluentIconProps | ISettingsIconProps`

packages/react-icons/template.js
Philip-Scott: Some icons can be filled iirc, is this controlled by a property on the SVG, or is it it's own separate icon?

packages/react-icons/package.json
```diff
@@ -0,0 +1,30 @@
...
+  "license": "MIT",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/microsoft/fluentui-system-icons.git"
+  },
+  "scripts": {
+    "clean": "shx rm -rf ./components",
+    "cleanSvg": "shx rm -rf ./intermediate",
+    "copy": "node ../../importer/generate.js --source=../../assets --dest=./intermediate --extension=svg --target=react",
+    "convert": "npx @svgr/cli --no-prettier --template template.js --expand-props none --typescript --replace-attr-values \"#212121={props.primary}\" -d ./components ./intermediate/",
```
Philip-Scott, line 15: Oh interesting, you are using SVGR. Just as an FYI, we had issues with this tool on a previous usage on the Fluid site due to groups on the SVG itself, but they were more complex shapes than icons. 

Do you have a way to view every icon once it's generated just to make sure svgr didn't cause any artifacts?

packages/react-icons/package.json
```diff
@@ -0,0 +1,30 @@
...
+    "unfill": "for file in ./components/*.tsx; do sed -i '.bak' 's/ fill=\"none\"//' $file; rm $file.bak; done",
+    "build": "npm run copy && npm run updatePrefix && npm run optimize && npm run convert && npm run unfill && npm run updateExport && npm run cleanSvg",
+    "prepublishOnly": "npm run clean && npm run cleanSvg && npm run build && npm run cleanSvg",
+    "postpublish": "npm run clean",
+    "updatePrefix": "for file in ./intermediate/ic*.svg; do mv \"$file\" \"${file/ic_fluent_/}\"; done",
+    "updateExport": "for file in ./components/*.tsx; do sed -i '.bak' 's/export const/export/' $file; rm $file.bak; done"
+  },
+  "devDependencies": {
+    "@svgr/cli": "^5.5.0",
+    "shx": "^0.3.2",
```
Philip-Scott, line 26: Is `shx` as a devDependency needed? It appears that it's only used as a wrapper on the clean commands and may be we can get rid of it. I'm not sure what terminal environments you run this so it could be fine, and if so please ignore this comment.

Philip-Scott (commented):
One question for clarification, are all icons are 24x24, with the 2px internal padding making the internal icons actually look more as 20x20, correct?

packages/react-icons/template.js
bheston: This is different than the model we discussed earlier, and while it may not be _too_ rigid, I wonder if it's not flexible enough. The interface we've been using that works with the flexibility of css is simply to apply an optional class name. Then any arbitrary style can be applied if necessary. The first reasonable example that comes to mind is something like `margin`, though that could potentially be applied generically to the `svg` element.

If we want to go this route, I think defaulting that value to `currentcolor` would be useful because much of the time, without knowing otherwise, an icon is the same color as neighboring text.

If this is the only way to provide the primary color, or eventually secondary or fill, etc. I wonder if this will be excess work to replicate on many icons. That is the values would always need to be known and passed to each icon.

Also, I'm not sure if this work with css-in-js where the colors might be determined at runtime.

This may be a good option for some consuming patters. I suggest maintaining className support as well.

packages/react-icons/package.json
```diff
@@ -0,0 +1,30 @@
...
+  "author": "Microsoft Inc.",
+  "license": "MIT",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/microsoft/fluentui-system-icons.git"
+  },
+  "scripts": {
+    "clean": "shx rm -rf ./components",
+    "cleanSvg": "shx rm -rf ./intermediate",
+    "copy": "node ../../importer/generate.js --source=../../assets --dest=./intermediate --extension=svg --target=react",
```
bheston, line 14: I didn't get a chance to experiment with this, but I was thinking the workflow should be to consume the optimized icons from the svg package. That way all logic about optimizing goes there and we can make any future updates in one place.

packages/react-icons/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "spnels-fluent-system-test0000",
```
bheston, line 2: @fluentui/react-icons

packages/react-icons/package.json
bheston: I used shx in the svg icons package to allow it to run on Windows. The official build is on Linux, but it was more convenient during dev time I thought.

packages/react-icons/package.json
chrisdholt: We should probably get an owner from Fluent UI React to ensure that the namespace works with their current expectations/requirements (if we haven't already).

packages/react-icons/template.js
chrisdholt: I think clarity on the prop contract here would be helpful. Is there a method to modify the HTML attributes on the SVG itself if needed? Consider a scenario where we need to apply ARIA, add a title attribute for some reason, etc...absolutely agree on className support as well. Are these intended to work only with the Fluent UI react implementation or with others *using* React? If the latter, I think we'll want to work through the prop contract to support a broader implementation.

packages/react-icons/package.json
spencer-nelson: Doing that will require that the SVG package be given any treatments needed for React. Are there any possible uses or properties that the React package will need that would be blocked by the optimizations for the SVG package? That's not my expertise, so I'll defer to others with more experience, but my first concern is the styling options, particularly as color is concerned. We have been asked to address possible multi-color icon scenarios. The way I see for that is to assign a color to individual paths, but all fills have been removed from the SVGs at the moment.

packages/react-icons/package.json
spencer-nelson: On that note, @bheston can you confirm which of these commands might not reasonably work for someone trying to maintain this on a Windows machine? I've always started my Windows dev boxes by installing gnu utils so don't recall which tools like sed/rm/whatever aren't readily available.

packages/react-icons/package.json
tomi-msft: Yeah, `@fluentui/react-icons` is the expected namespace for this package

packages/react-icons/template.js
spencer-nelson: They are expected to work for any React package. Assuming we define a package interface akin to:
```
export interface IFluentIconsProps {
	primary?: string;
	className?: string;
}
```
does this allow further extension to support other HTML attributes, either by product consumers or as a wider package implementation in the future?

spencer-nelson: Further expounding on the className, is the following worth pushing as a V1 option, with the interface defined above?
```
import * as React from "react";
import { JSX } from "react-jsx";
import { IFluentIconsProps } from "./IFluentIconsProps.types";

const AccessTime24Regular = (
  iconProps: IFluentIconsProps,
  props: React.HTMLAttributes<HTMLElement>
): JSX.Element => {
  const { primaryFill = "#212121", className } = iconProps;
  return (
    <svg
      width={24}
      height={24}
     
      xmlns="http://www.w3.org/2000/svg"
      className={className}
      {...props}
    >
      <path
        d="<snip/>"
        fill={primaryFill}
      />
      <path
        d="<snip/>"
        fill={primaryFill}
      />
    </svg>
  );
};

export default AccessTime24Regular;
```
It was earlier mentioned to me that we might not want to spread all the props in this way, but I don't know another way to get the HTML attributes in.

packages/react-icons/package.json
spencer-nelson: Thanks. We don't have a visual regression test set up at the moment to test for that, but given the transforms happening and the simple svgs we build in the first place, we won't have any groups. If those were the breaking elements for you I feel we'll be good to go with svgr, but what did you all settle on using and how difficult would it be to swap in for svgr here?

Philip-Scott: I ended up opening the SVG in Inkscape, and ungrouping everything to fix the issue and that worked. Also validated that if our designers gave us SVGs without being grouped, there were no artifacts. This was late 2019 so maybe the issue was fixed, but I thought it was worth mentioning as an FYI :)

spencer-nelson:
> One question for clarification, are all icons are 24x24, with the 2px internal padding making the internal icons actually look more as 20x20, correct?

That is correct. They will visually appear as 20x20 and should be placed 2px offset up and to the left to get them pixel-perfect placement compared to systems without the 2px internal affordance.

packages/react-icons/package.json
Philip-Scott: Due to the open source nature of the icons package, it sounds fine to use to allow someone without the gnu utils to clone and use fully the repo. Thanks for the explanation Brian. 

In that case you are using `rm` and `mv` on the `updatePrefix` and `updateExport` scripts without shx, so those commands may not work on Windows :)

packages/react-icons/package.json
bheston: It seems we'd need/want to support the same color capabilities across all published formats, so I don't see any reason React would need something that we were removing in svg. If anything it allows us to standardize the way we treat the icons, for instance anywhere we support a secondary color or fill it would work the same way at the actual svg level. The interface might vary based on implementation.

packages/react-icons/template.js
```diff
@@ -0,0 +1,27 @@
...
+	exports.declaration.name = componentName.name
+
+	return tpl.ast`
+		import * as React from "react"
+		import { JSX } from "react-jsx"
+
+		import { IFluentIconsProps } from '../IFluentIconsProps.types'
+
+		const ${componentName} = (iconProps: IFluentIconsProps, props: React.HTMLAttributes<HTMLElement>) : JSX.Element=> {
+		const { primaryFill = "#212121", className } = iconProps;
```
bheston, line 19: I think the fill needs to default to null for this to work correctly. Many usages won't want to specify a color for each icon at this level. It doesn't account for the separation of the icon as content and the styling as css. We can author the icons with 212121 but there isn't really a default that seems convenient here aside from "currentcolor", which might be undesirable for other uses. The upside here is that null fill everywhere renders as black by default.

packages/react-icons/package.json
spencer-nelson: In digging around and playing with it, I can't find a way to make it work well on both systems. We either hit issues with wildcard expansion or the variations on looping through the files break on one system or the other. Given my lack of time I'm going to suggest we instead tell Windows devs that want to copy/maintain the pipeline use WSL and we will try to just stick to a Ubuntu-friendly solution as that's what the GitHub workflows use to run them anyway.

tomi-msft (approved):

packages/react-icons/package.json
spencer-nelson: Given that we don't want to block this, I'll add an issue to handle multi-color options and specifically mention tightening up on this flow.

PR #186, opened 2021-02-23
Add publish steps for @fluentui/react-icons

spencer-nelson:


spencer-nelson:
After playing around with things, it's simplest to just change the svgr usage from npx to a node script. The successful pr validate run is:
https://github.com/microsoft/fluentui-system-icons/actions/runs/594658003
so I'm going to move the actual publishing step to the publish branch.

tomi-msft (approved):

PR #208, opened 2021-03-09
Remove "preserve-vector-representation" option from generated iOS image assets

cy255:
Description
Remove "preserve-vector-representation" when generating iOS image assets to 
1. Reduce bundle size impact (vector data preserved assets takes up more space).
2. Each icon is created for a specific size so stretching it to another size is not recommended.

Impact
This change impacts how iOS image assets are generated


importer/process_ios_assets.py
```diff
@@ -89,8 +89,7 @@ def process_assets():
                     "author": "xcode"
                 },
                 "properties": {
-                    "template-rendering-intent": rendering_intent,
-                    "preserves-vector-representation": True
```
isatoderici, line 93: Don't we wanna keep the property, but set it to False?

nickromano (approved):
Nice! I didn't realize this had an impact on bundle size. You're right that we don't need it with our size-specific icons.

Thank you for contributing!

importer/process_ios_assets.py
cy255: The default value of the "preserves-vector-representation" is false so not necessary to make it explicit.

PR #213, opened 2021-03-19
update example build script to remove mention of react-icons-preview

tomi-msft:
This PR is to change the example package to be able to show the icons without running the create script. Users should just be able to see the icons they have available in a webpage without going through extra hoops.


spencer-nelson:
If we have pre-generated files we will need to generate them as part of the publish script so they are kept up to date.

spencer-nelson (approved):

PR #238, opened 2021-05-01
Add commonjs module to package.json

tomi-msft:
This PR is to add CommonJS modules to the build for cases  where the library is used within a Server side rendering environment, which generally uses Node and hence might require a CJS counterpart of the library.

### DIFF
--- a/packages/react-icons/package.json
+++ b/packages/react-icons/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@fluentui/react-icons",
   "version": "1.1.122",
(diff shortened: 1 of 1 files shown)


willhou (approved):

spencer-nelson:
@tomi-msft do we need to add anything to the build pipeline/delivery to make this readily available for people?

tomi-msft:
@spencer-nelson adding it to the package.json and updating the build script there should be all that we need to do

PR #250, opened 2021-05-12
Update template to bundle filled and regular states

tomi-msft:
This PR is to update the template for react-icons to bundle the filled and regular states for each icon. It also defaults the fill to `currentColor`, so that the icon can inherit its container color


spencer-nelson (changes requested):
I don't think we want all icons to always have the data for both Filled and Regular, as this doubles the size of every component used, regardless of whether both Filled and Regular are needed. Would it be better to let users decide if they wanted to optimize for size or for ease of coding?

Jahnp:
@spencer-nelson, @tomi-msft and I were chatting about this yesterday and I agree completely with the concern around component sizes. The issue as I understand it is that many Fluent vNext experiences using these icons will need these interactive states, so having the `filled` toggle would be a big convenience for folks needing that.

Just brainstorming here, I wonder if it'd be too kludgy to offer a "bundled" version that includes `filled` for convenience, and leave the non-filled version. Or, maybe there could be an additional utility function/component that allows easy flippability between 2 given icons. At minimum a code example with basic boilerplate logic should be provided since it sounds like this'll be a really common scenario.

Thoughts?

tomi-msft:
@Jahnp I'm a little leery with adding another icon package to the 4 that Fluent already offers. The way it stands right now, we have `font-icons-mdl2`, `react-icons-mdl2`, `react-icons-mdl2-branded`, and this package, `react-icons`. I'm concerned that adding another icon package would increase confusion as to when to use which package where.

Jahnp:
@tomi-msft for sure, I actually wasn't thinking of another package for that exact reason—more like a different import path in the same package.

Here's the gist of what I had in mind (though "bundled" definitely doesn't sound right):
```tsx
import { AccessTime24Filled } from "@fluentui/react-icons/bundled";
```

That definitely feels weird, and I'm not sure there's a pattern for this kind of thing in other packages. It also **could** triple the npm package size.

Another idea would be to offer a simple utility to swap any two icons based on some condition. Maybe `wrapIcon` could be extended to do this (though I think that's probably the wrong tool).

Here's what that could look like:
```tsx
import { AccessTime24, AccessTime24Filled, swapIcon } from "@fluentui/react-icons";

const SwappedAccessTime24Filled = swapIcon(<AccessTime24 />, <AccessTime24Filled />);

const rootElement = document.getElementById("root");
ReactDOM.render(
    <div>
        <AccessTime24Filled iconProps={...iconStyleProps}  />
        <SwappedAccessTime24Filled swapped={someCondition} />
    </div>, 
    rootElement
    )
```

Thoughts? Is that too silly of a utility?

spencer-nelson:
There are some that never use the Filled icons, as they just change background states or borders for something that is/isn't selected. I see the utility, but would prefer to default to the separated out versions so we don't stop someone from that level of optimisation. I would argue that a combined component would be in the `@fluentui/react-icons/combined` (or whatever your choice of words) and `@fluentui/react-icons` itself would be the vanilla.

Another way this issue comes up is a combined size option so that one just imported the icon by name a la
`import { AccessTime } from "@fluentui/react-icons/<???>`
and then uses
`<AccessTime size=12 theme=filled/>`

Jahnp:
Oh interesting, I hadn't thought about the size options too.

Your proposal of controlling both size & filled as props is definitely elegant! But I can see how combining size + filled/unfilled like that could result in even larger imports, and I worry there could be a lot of SVG data that might go unused, especially since I *think* it'd be unusual for someone to use multiple sizes of a single icon on a given page? If you're using `AccessTime12`, it's unlikely you'd also need to use `AccessTime24` in the same UX, right? 

+1 to keeping the standalone icons in `@fluentui/react-icons` though.

I'm curious @spencer-nelson, do you have thoughts on the swap utility direction? It'd be more work for consumers to set them up, but does allow more flexibility/composability. I could imagine consumers creating their own utility to do something like this if the `combined` option weren't feasible for them.

tomi-msft:
With the swap utility, it essentially would be a function that creates a component that is the opposite state of the imported one? I guess, why would they not just import the other Icon?

Jahnp:
@tomi-msft yep, it's essentially that, good point.

And now that I think about it, the swapping could probably be handled with a simple inline conditional/ternary check. Bah, probably not helpful unless those components are the output we'd be providing in the `combined` option!

spencer-nelson:
The multiple size option would be for folks wanting reactive layouts that include information density changes, but you're right that it quickly balloons the bundle size. It also opens up possible updating problems for an application that does load/perf testing on one version, then updates to a later version where we've added more sizes.

I'd be curious to hear the User Research story behind how often both Regular and Filled icons are used in the same layouts such that we need to add this for folks. I doubt it exists as such, but it's worth asking around for designers that build for that.

Jahnp:
Thanks for the context @spencer-nelson , density changes is a scenario makes sense for the size axis. 

@tomi-msft, just curious if you have thoughts now on how you might want to proceed here? Or if there's any way I can help move the conversation forward?

On the product side, I've learned there are some design directions being considered that would use both the filled/unfilled states which would benefit from this option being available. I'm happy to follow up or connect dots if it'd be useful to you. Let me know!

tomi-msft:
Hi @Jahnp, I definitely see the need to be able to bundle the filled and regular icons, and provide a way to easily toggle the two states. I think moving forward the best way to go would be to create a utility that will bundle the filled and regular icons. I think it can be something very simple, very similar to what you outlined earlier. 

I think we can go with the filled prop dynamically rendering which icon should be passed in, or we can render both icons on the page and use css to toggle which icon gets displayed on the page, and set the opacity of the other to zero. The css option is more performant, as styling using css is always faster than doing it programatically

Jahnp:
Hey @tomi-msft, that's great! I've spoken to a few folks on my team and confirmed that providing a way to do this will actually encourage uptake of the SVG icons; one engineer was concerned about writing boilerplate code to do the swapping and claimed the cost of doing so would keep him from switching from font-based icons to these.

To be clear about the utility, do you mean a runtime utility similar to my `swapIcon` utility idea above? Or a build-time utility that creates combined versions in a different path? I think either would be fine, though the runtime utility would keep the package size down.

> ...or we can render both icons on the page and use css to toggle which icon gets displayed on the page, and set the opacity of the other to zero. The css option is more performant, as styling using css is always faster than doing it programatically

This sounds fair to me, the one thing I'd suggest is to consider using `display: none` or `visibility: hidden` instead of `opacity: 0` as I believe many screen readers will announce both icons with `opacity: 0`. The first two properties should mask the hidden icon from screen readers.

PR #261, opened 2021-05-25
Use one imageset instead of two for LTR and RTL icons.

isatoderici:
In the current implementation we use one imageset for LTR and a separate one for RTL. The scope of this change is to use just one imageset for both LTR and RTL directions.

The icons were also renamed, so instead of "ic_fluent_arrow_export_ltr_20_filled" and "ic_fluent_arrow_export_rtl_20_filled", we have just "ic_fluent_arrow_export_20_filled".


isatoderici:
Implemented by @nickromano !

PR #264, opened 2021-05-26
Update locale case

nickromano:


importer/process_ios_assets.py
```diff
@@ -68,10 +68,23 @@ def add_localized_set(lang_locs, original_icon_names, icon_assets_path):
             imageset_contents_path = os.path.join(imageset_path, "Contents.json")
             contents_json = json.load(open(imageset_contents_path))
 
+            # Xcode is specific about the capitalization used for locales.
+            # "bg" -> "bg"
+            # "bg-bg" -> "bg-BG"
+            # "sr-latn" -> "sr-Latn"
+            locale_components = lang_loc.split("-")
```
spencer-nelson, line 75: How does this work for something like ja-jp-jp? Should it be extended to treating all past the language bit with this capitalization treatment?

nickromano: Good point, I was meaning to ask about that I’m not sure Xcode uses the third part for their localization. Do you think we should drop those?

spencer-nelson: I'd prefer to keep them if Xcode will use them. We don't have any asks for that level of localization now, so what is in now works. I was just trying to be pro-active.

spencer-nelson (approved):

importer/process_ios_assets.py
nickromano: Cool we can merge for now but it does look like it throws warnings for the icons with three parts. Maybe in a follow up we can just drop those icons on iOS

Issue #240, opened 2021-05-02
Gradle build error


tipapro:
Execution failed for task ':app:dataBindingMergeDependencyArtifactsDebug'.
> Could not resolve all files for configuration ':app:debugCompileClasspath'.
   > Could not find :unspecified:.
     Required by:
         project :app > com.microsoft.design:fluent-system-icons:1.1.122
         

Maybe this is caused by '<artifactId>unspecified</artifactId>' line?

MichaelC11:
Hi there! I hit this issue too :(

Note: I did notice that an earlier version (e.g., 1.1.112) doesn't have this issue - so I wonder if it's a more recent regression?

tipapro:
> Hi there! I hit this issue too :(
> 
> Note: I did notice that an earlier version (e.g., 1.1.112) doesn't have this issue - so I wonder if it's a more recent regression?

I solve it with:
    implementation ("com.microsoft.design:fluent-system-icons:1.1.122") {
        exclude module: 'unspecified'
    }

MichaelC11:
Neat, thanks!! :)

yunu7067:
> > Hi there! I hit this issue too :(
> > Note: I did notice that an earlier version (e.g., 1.1.112) doesn't have this issue - so I wonder if it's a more recent regression?
> 
> I solved it with:
> implementation ("com.microsoft.design:fluent-system-icons:1.1.122") {
> exclude module: 'unspecified'
> }

It works! Tnx!!

Adamantcheese:
This has been occurring since 117; 116 is the last version that doesn't have this issue. Seems to be related to the move to Maven Central from JCenter in regards to that repository's deprecation.

willhou:
@tipapro @MichaelC11 @yunu7067 @Adamantcheese 

Sorry about that. Can you try adding `@aar` to the end of the dependency? This should explicitly request for the aar hosted on maven central instead of a jar(by default).
```groovy
implementation 'com.microsoft.design:fluent-system-icons:1.1.127@aar'
```

PR #277, opened 2021-06-10
Removing 'unsepecified' dependencies from maven pom file

tito0224:
When the package is published to maven the pom file that is generated contains a list of dependencies. However since this package only specified a local lib folder and wasn't filtering it during packaging it was being added to the pom file as "unspecified":

```xml
<dependencies>
    <dependency>
      <groupId/>
      <artifactId>unspecified</artifactId>
      <version/>
    </dependency>
  </dependencies>
```

Since there are no dependencies either local libs or maven ones I removed the dependency block so that the resulting pomfile will look like this:

```xml
<dependencies />
```

### DIFF
--- a/android/library/build.gradle
+++ b/android/library/build.gradle
@@ -20,8 +20,4 @@ android {
 
 }
 
-dependencies {
-    implementation fileTree(dir: 'libs', include: ['*.jar'])
-}
-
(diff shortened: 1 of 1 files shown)


willhou:
Thanks, related to #240

willhou (approved):

Issue #276, opened 2021-06-10
React icons wrap svg on span making it change size according to line-height


PatoBeltran:
Hi there!

I'm trying to migrate my product to these icons following https://hackmd.io/RPZl3VIDRz-sobbE48IJuQ ... while doing so I noticed that the SVG is always wrapped in a `<Span />` element, which ends up making the icon bigger than expected.

I noticed that in my case, the line height is set to 1.15, which makes the span height `24px`  even though the SVG itself is `20px` height. If that span had `line-height: 0` set (or `display: inline-flex` to make it ignore the line height), then the size of the SVG would be respected and there wouldn't be extra growth.

This extra whitespace makes it so that my icons end up misaligned whenever I align them using flex (e.g. `justify-content: center`).

tomi-msft:
Hi @PatoBeltran , thanks for filing the issue! I can go ahead and add default styling to the span around the svgs that will set `display: inline-flex` and that should fix your issue!

sjwilczynski:
Hi, I wanted to bump to version 1.1.134 but even then I don't see the default stiles for the icon. The `root-span` class seems to be empty:
![image](https://user-images.githubusercontent.com/19153565/124600883-7f14cf00-de67-11eb-9a6b-7b103daab35b.png)
![image](https://user-images.githubusercontent.com/19153565/124600900-8340ec80-de67-11eb-9835-98b06183defb.png)
@tomi-msft do you know what is causing the issue?

tomi-msft:
This should be fixed now

PR #282, opened 2021-06-17
Create bundleIcon utility to bundle regular and filled states

tomi-msft:
This PR adds the `bundleIcon` utility to the react-icons package, as well as adds default styling to the regular icons, to address issue #276. Finally, it adds the `sideEffects: false` flag to the `package.json` to enable treeshaking and dramatically speed up the build time for the tsc module.


packages/react-icons/src/utils/BundledIcon.scss
```diff
@@ -0,0 +1,10 @@
+.bundled-icon-span .root-span.filled {
+    display: none;
+}
+.bundled-icon-span:hover .root-span.regular, .bundled-icon-span:active .root-span.regular .bundled-icon-span:focus .root-span.regular{
```
Jahnp, line 4: `:hover` is a sensible default for toggling filled/unfilled states, but I still believe there will be lots of cases where we'd want to customize that behavior, aka making the bundled component controllable.

For example, in a command bar or left nav, those states might change based on user selection rather than hover. In ODSP we will initially only be changing color, not filled state, though that may change in time.

A better scenario would be disabling toggling if the icon is in a selected nav item. If an icon is already filled, it's probably undesirable to switch back to unfilled on hover.

Do you have concerns about adding that capability in general? It could be added to a future iteration if you'd rather proceed with current behavior for now. I just worry it'll be an adoption blocker if users need to write extra code to override default behavior.

packages/react-icons/README.md
```diff
@@ -21,21 +21,41 @@ In order to use these icons, simply import them as `import { [Componentname][siz
...
 import * as ReactDOM from "react-dom";
-import { AccessTime24Filled } from "@fluentui/react-icons";
+import { AccessTime24Filled, AccessTime24Regular, bundleIcon } from "@fluentui/react-icons";
+import './Test.scss';
 
 const iconStyleProps: IFluentIconsProps = {
     primaryFill: "purple",
     className: "iconClass"
 };
+const BundledIcon = bundleIcon(AccessTime24Filled, AccessTime24Regular)
```
spencer-nelson, line 52: This is super nitpicky, but in a 'real world' scenario this should be called AccessTime, not BundledIcon, I would think.

tomi-msft: Its been changed!

spencer-nelson (approved):

PR #314, opened 2021-08-31
remove sourceMaps and unnecessary files from project output

tomi-msft:
This PR is to update the react-icons package to only publish the lib folder, and remove sourceMaps from production build. Currently the react-icons package publishes the src folder and has a sourceMap for every icon, which creates a lot of unnecessary files and massively slows down download speed. This reduces the number of files needed to download to ~30% of the previous number of files


spencer-nelson (approved):

digarza:
This PR might have removed the type mapping files in the lib folder, was that intentional? Currently, I'm having to ignore compiler errors from not finding declaration files in the latest version

PR #320, opened 2021-09-11
Consolidate typescript definition files for esm and cjs modules

tomi-msft:
This PR consolidates the typescript definition files in the `lib/esm` folder and the `lib/cjs` folders into one `index.d.ts` file to reduce the number of files emitted by the react-icons package. This significantly reduces the total number of files emitted, which will reduce the time to download the package.


packages/react-icons/generateRollup.js
```diff
@@ -0,0 +1,59 @@
...
+  `declare module '@fluentui/react-icons' {
+  export * from '${rootInPackage}/esm/index';
+}`,
+];
+
+// make a module declaration for each .d.ts file (and delete the original)
+for (const outType of outTypes) {
+  // This gives a list of relative paths like 'esm/components/Accessibility16Filled.d.ts'
+  const files = glob.sync(`${outType}/**/*.d.ts`, { cwd: outDir });
+  console.log(files);
```
spencer-nelson, line 34: Do we still need this console.log? In general it will have a complete set of the files, so won't usually be useful but will still take up space and time logging things.

tomi-msft: Good point, will remove that!

packages/react-icons/generateRollup.js
```diff
@@ -0,0 +1,59 @@
...
+// https://github.com/jprichardson/node-fs-extra/blob/master/docs
+const fs = require('fs-extra');
+// https://www.npmjs.com/package/glob
+const glob = require('glob');
+
+const outDir = path.join(__dirname, '/lib');
+// NOTE: esm comes first to increase chances of it being suggested first in auto-imports
+const outTypes = ['esm', 'cjs'];
+
+// // copy test output to fake react-icons package
```
spencer-nelson, line 12: Is there long-term value in leaving these commented lines in?

tomi-msft: I'll be removing the test comments on line 12, as well as the commented out code block. I think the comment on line 9 is useful

spencer-nelson (approved):

PR #325, opened 2021-09-30
Major bump react-icons and release react-icons-sized

tomi-msft:
This PR major bumps the react-icons package to 2.x and creates the react-icons-sized package. The react-icons package will now have only one component per icon, and each icon will be resizeable using regular styling(i.e setting `height` and `width` props, setting `fonstSize` prop, etc). 

Another change is to the way the package is structured, to make deep imports easier. The `components` folder has been removed, so users will be able to import icons like so: `import { AccessTimeRegular } from "@fluentui/react-icons/lib/AccessTimeRegular"`. This will drastically decrease render time of these icons on a screen.

Finally, the README has been updated with the major changes, and what that means for users going forward.

There are a lot of files in the PR, but most of them are the emitted icon files from two packages. Here are the main files to check. Note that there are two of them for each, since there is one for each package:

- convert.js
- README.md
- generateRollup.js
- package.json
- tsconfig.json


layershifter:
@tomi-msft can you please use `@fluentui/react-make-styles` as `peerDependency` + `devDependency`?

⬇⬇⬇

```diff
- "dependencies": {
-   "@fluentui/react-make-styles": "^9.0.0-alpha.70"
+ "peerDependencies": {
+   "@fluentui/react-make-styles": "^9.0.0-alpha.70"
  },
```

---

### Why?

In Fluent UI we currently use fixed versions without carets: https://github.com/microsoft/fluentui/blob/32414b738c31ad29cb67af3cb9d6e77929c18c55/packages/react-components/package.json#L47

(_see microsoft/fluentui#19748 for more details_)

Mixing of carets and fixed versions can potentially lead to duplicated versions of `@fluentui/react-make-styles` that will cause serious problems for customers as there should be single instance of `makeStyles()`.

packages/react-icons/package.json
```diff
@@ -35,8 +36,14 @@
     "typescript": "~4.1.0",
     "yargs": "^14.0.0"
   },
+  "peerDependencies": {
+    "@fluentui/react-make-styles": "9.0.0-alpha.70"
```
layershifter, line 40:
```suggestion
    "@fluentui/react-make-styles": "^9.0.0-alpha.70"
```

This should still use caret otherwise it will be the same problem as with `dependencies`

PR #328, opened 2021-10-04
Add a package which builds SVG sprites

thure:
Packages which depend directly on `@fluentui/react-icons` can cause very lengthy build times, since some bundler configurations will cause the bundler to iterate through each of the 8,833 icons in this project. This ~is~ was true at least for Storybook consuming `@fluentui/react-components`.

The situation calls for a more comprehensive solution than even the one contained in this PR. I think the ideal solution for a React stack would be to offer a module which tracks which icons are used by an app (maybe similar to the `makeStyles`/`useStyles` pattern), and render either a one-off sprite during development or a sprite optimized for exactly the icons used in the app in production.

As a first step, though, the spriting solution in this PR offers these benefits:
- Developers don’t need to install this package as a dependency
- The user agent can cache requests for sprites
- Icon assets no longer affect project bundle size

SVG sprites have been well supported by user agents [for more than 5 years by now](https://caniuse.com/mdn-svg_elements_symbol). This package uses [`svgstore`](https://github.com/svgstore/svgstore) to generate the sprites.


packages/svg-sprites/svgo.config.yaml
```diff
@@ -0,0 +1,5 @@
+plugins:
+  - removeViewBox: false
```
bheston, line 2: Can `precision` go in here as well?

packages/svg-sprites/unfill.js
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) Microsoft Corporation.
```
bheston, line 1: Can we have a unified treatment for icons, at least within svg type? If I recall the `none` fill comes from the root `svg` element, which is beneficial to remove for size in all cases. The `212121` fill comes from the `path`. In the original package we opted to specify no fill to remove any _opinion_ on color and allow consumers to (likely) set to `currentcolor` in their style sheets. I suggest going forward with that since it would work in this new package, and would also make the download size smaller. I do agree that `currentcolor` is probably very common, but since it's not a given I would lean toward being more concise.

Can we share the file from the other package or move it to a utilities directory?

bheston (changes requested):
I wasn't familiar with that. Nice optimization for those who could use it.

packages/svg-sprites/svgo.config.yaml
thure: I believe so, yes!

packages/svg-sprites/unfill.js
thure: I'll look into that, thanks for the feedback.

packages/svg-sprites/svgo.config.yaml
thure: Is there a standard value for `precision` we should use here?

thure:
@bheston I've given this another stab informed by your feedback, and I've published and tested the result. As far as I can tell, everything works as expected, and now precision is correct and there're no superfluous `fill` attributes ✨

thure:
I am also considering a follow-up where this package can be added as a developer dependency, then used to select specific icons to add to a single sprite for a project to use.

A comparison with that idea and this PR is here: https://github.com/thure/fluentui-system-icons/pull/1/files

packages/svg-sprites/package.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "@fluentui/svg-icon-sprites",
```
bheston, line 2: Probably rename the folder for consistency.

bheston (approved):

packages/svg-sprites/package.json
tomi-msft: Maybe `"@fluentui/react-svg-icon-sprites"`

thure: 'fluent' seems redundant since it's already scoped to @fluentui, and the adjacent package is named `@fluentui/svg-icons`.

Maybe `@fluentui/svg-sprites`? I'll go with that name and adjust the directory name for now, let me know if it should change.

thure: Ah, and that happens to be the name of the directory already.

thure:
Looks like it just needs a review from someone with write access… @bheston last time your approval wasn't enough to merge.

packages/svg-sprites/README.md
```diff
@@ -0,0 +1,40 @@
...
+For Storybook, `middleware.js` allows the configuration of `http-proxy-middleware` like so:
+
+```js
+const { createProxyMiddleware } = require('http-proxy-middleware')
+
+const expressMiddleWare = (router) => {
+  router.use(
+    '/sprites',
+    createProxyMiddleware({
+      target: 'http://cdn.jsdelivr.net/npm/fluentui-svg-icon-sprites@1.1.163/',
```
spencer-nelson, line 18: Does this reference need to change with the package name change?

thure: Ah, indeed, thanks for catching this.

vnbaaij:
As this is enabling to work with sprites, I was wondering if it would make sense to 'pack' all the variants of an icon in 1 sprite. The size of the individual icons is so small that even a packed version is still a very small download (largest one is ~20k).
The way this can be achieved in `package.json` is to replace the `sprites` script with the following :
```
"sprites": "mkdir -p ./sprites && for j in ./icons/**; do svgstore  -o \"./sprites/$(basename $j)/$(basename $j).sprite.svg\" \"$j/*.svg\"; done",
```

packages/svg-sprites/package.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "@fluentui/svg-sprites",
+  "version": "1.1.164",
```
tomi-msft, line 3: Update version to 1.1.167 to be in line with other icon packages

packages/svg-sprites/package.json
bheston: I was suggesting the package and directory names should both be `svg-icon-sprites`. By itself I don't think `svg-sprites` alone sufficiently identifies them as icons. It could be any svg content, but I think the consistency in naming between all modern `icon` packages is beneficial.

bheston:
> Looks like it just needs a review from someone with write access… @bheston last time your approval wasn't enough to merge.

I believe @spencer-nelson was able to push through one of my previous PRs.

tomi-msft (approved):

PR #329, opened 2021-10-05
Update publish.yml

tomi-msft:
Updates publish.yml to be able to bump individual packages

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -28,6 +28,15 @@ jobs:
         import os
         current_version = sys.stdin.read().strip().split(\"'\")[1]
         major, minor, patch = current_version.split('.')
+        react_major = {int(major) + 1}
+        react_minor = {minor}
+        ios_major = {major}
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -112,7 +121,7 @@ jobs:
     # Needs to be "-E" instead of "-r" on macOS
     - name: Replace version numbers in ios/README.md
       run: |
-        sed -i.bk -r "s/[0-9]+\.[0-9]+\.[0-9]+/$NEW_VERSION/g" ios/README.md
+        sed -i.bk -r "s/[0-9]+\.[0-9]+\.[0-9]+/$IOS_VERSION/g" ios/README.md
```
nickromano, line 124: Would you mind reverting this one back to `NEW_VERSION` or changing the rest of the entries in this file to `IOS_VERSION` and `ANDROID_VERSION` or maybe `NATIVE_VERSION` since both of those are kept in sync?

PR #332, opened 2021-10-11
Use SVGs for iOS image sets

cy255:
Description
SVGs and PDFs without vector representation provide similar performance (size impact and runtime performance) so changing iOS image set default file format to SVGs instead of PDFs for consistency across platform. A detailed analysis on iOS image file format can be found here: https://microsoft.github.io/apple-ux-guide/ 

VERIFICATION
-Running `npm run deploy:ios` locally did generate imagesets containing SVG images.

Impact
This changes the image file format for iOS image assets from PDF to SVG.


anandrajeswaran:
To "stage" this change, thoughts on having it as a parameter and having it be off by default to start with? When we have had a few clients adopt it and confirm they are seeing similar data to what we've seen in pdf vs. svg trials we can switch it to be the default?

PR #353, opened 2021-11-08
Fixes for Android version number issues when other packages fail to publish

spencer-nelson:


spencer-nelson:
The primary change here is to push Android to the last platform to publish, as the Sonatype publishing doesn't allow re-publishing the same version, or pulling down an 'old' one.
Also includes a perf change to how the Android library is built, replacing find...replace to just batching replace

nickromano (approved):

PR #360, opened 2021-11-18
Add babel-make-styles build

tomi-msft:
This PR adds a build time transform using `babel` and `babel-make-styles` to the build script in order to reduce the bundle size hit of using make-styles when the package is released to npm. It also makes some style changes to the icon, and moves the constants used in `bundleIcon` to their own file

### DIFF
--- a/packages/react-icons/src/utils/constants.tsx
+++ b/packages/react-icons/src/utils/constants.tsx
@@ -1,0 +1,2 @@
+export const FILLED_CLASSNAME = "fui-Icon-filled";
+export const REGULAR_CLASSNAME = "fui-Icon-regular";
--- a/packages/react-icons/convert.js
+++ b/packages/react-icons/convert.js
@@ -52,6 +52,7 @@ function processFiles(src, dest) {
   indexContents += '\nexport { default as wrapIcon } from \'./utils/wrapIcon\''
   indexContents += '\nexport { default as bundleIcon } from \'./utils/bundleIcon\''
   indexContents += '\nexport * from \'./utils/useIconState\''
+  indexContents += '\nexport * from \'./utils/constants\''
   fs.writeFileSync(indexPath, indexContents, (err) => {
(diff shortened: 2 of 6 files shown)


packages/react-icons/src/utils/constants.tsx
```diff
@@ -0,0 +1,2 @@
+export const FILLED_CLASSNAME = "fui-Icon-filled";
```
spencer-nelson, line 1: is there a particular reason this was changed from fluentui-icon-xxx to fui-Icon-xxx?

spencer-nelson: and are there any particular knock-ons to worry about? Presumably the classnames would have been already used/referenced, so is this not breaking?

spencer-nelson (changes requested):

packages/react-icons/src/utils/constants.tsx
tomi-msft: this was changed to be inline with other fluentui component classnames naming convention. This is not breaking since this will get converted anyway by the babel-plugin

spencer-nelson (approved):

packages/react-icons/src/utils/constants.tsx
```diff
@@ -0,0 +1,2 @@
...
+export const REGULAR_CLASSNAME = "fui-Icon-regular";
```
layershifter, line 2: I forgot to mention this, but let also use the same naming:

```tsx
export const iconFilledClassName = "fui-Icon-filled";
export const iconRegularClassName = "fui-Icon-regular";
```

https://github.com/microsoft/fluentui/blob/42c69701f6af74ab511549343c01a88d1889cfd3/packages/react-card/src/components/Card/useCardStyles.ts#L5

layershifter:
Verified the output, works like a charm 💎

image

PR #364, opened 2021-11-22
@fluentui/react-icons: update FluentIconsProps

layershifter:
This PR adds return type to `useIconState()` hook to simplify generated TypeScript definitions.

### Before

image

### After

image

---

Because of emitting `<reference types="react" />`it creates a lot of noise for [API Extractor](https://api-extractor.com/):

image

https://github.com/microsoft/fluentui/pull/20563

### DIFF
--- a/packages/react-icons/package.json
+++ b/packages/react-icons/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@fluentui/react-icons",
-  "version": "2.0.151-beta.2",
+  "version": "2.0.151-beta.3",
   "sideEffects": false,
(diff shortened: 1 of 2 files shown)


packages/react-icons/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@fluentui/react-icons",
-  "version": "2.0.151-beta.2",
+  "version": "2.0.151-beta.3",
```
layershifter, line 3: @tomi-msft mentioned that it's required to bump version to publish

tomi-msft (approved):

PR #373, opened 2021-12-06
chore(react-icons): Do not minimize build output and chunk icon files

ling1726:
Webpack builds a lot slower due when using running on large minified
files. To fix this, the PR adds a `.babelrc` file with `compact: false`
so the result is kept prettified (which is the default output of
Typescript).

Additionally, modified `convert.js` so that the output icons are
chunked. This means that there will only be 500 icons per file. All
contents of chunked files are exported from `index.ts`. This improves
the build to ~1.5 minutes.


tomi-msft (approved):
lgtm!

ling1726:
addressed by acorjs in acornjs/acorn#1083

PR #404, opened 2022-02-10
Styling update

tomi-msft:
This PR adds react as a peer dependency and adds default styling to set `forced-color-adjust: auto`


packages/react-icons/package.json
```diff
@@ -40,7 +40,8 @@
     "yargs": "^14.0.0"
   },
   "peerDependencies": {
-    "@griffel/react": "^1.0.0"
+    "@griffel/react": "^1.0.0",
+    "react": "^17.0.1"
```
layershifter, line 44: This should match Fluent's range:

```suggestion
    "react": ">=16.8.0 <18.0.0",
```

https://github.com/microsoft/fluentui/blob/a99218c3e02097545901e6a56cd6fe8aaae83661/packages/react-avatar/package.json#L56

packages/react-icons/src/utils/useIconState.tsx
```diff
@@ -6,7 +6,9 @@ const useRootStyles = makeStyles({
         display: 'inline',
         lineHeight: 0,
 
-        "@media screen and (-ms-high-contrast: black-on-white)": {fill: 'windowtext'}
+        "@media (forced-colors: active)": {
```
ling1726, line 9: I think we could be more specific and only target the `svg` element here. It's the actual element that has the color inherit issue

https://github.com/w3c/csswg-drafts/issues/6310
https://bugs.chromium.org/p/chromium/issues/detail?id=1242706

layershifter (dismissed):

PR #416, opened 2022-03-10
Latest version update

tomi-msft:
This is the first part of 2 PRs to start releasing the `react-icons` package as a non beta release. To align with other converged FluentUI components, it will have the rc tag at the end of it.


.github/workflows/publish.yml
```diff
@@ -110,9 +110,21 @@ jobs:
     # Needs to be "-E" instead of "-r" on macOS
     - name: Replace version number in react-icons/package.json
       run: |
-        sed -i.bk -r "s/\"version\": \"[0-9]+\.[0-9]+\.[0-9]+(-beta\.[0-9]+)?\"/\"version\": \"$REACT_VERSION\"/g" packages/react-icons/package.json
+        sed -i.bk -r "s/\"version\": \"[0-9]+\.[0-9]+\.[0-9]+(-rc\.[0-9]+)?\"/\"version\": \"$REACT_VERSION\"/g" packages/react-icons/package.json
         rm packages/react-icons/package.json.bk
 
+    # Needs to be "-E" instead of "-r" on macOS
```
spencer-nelson, line 116: This isn't needed any more, is it? That manifest was removed.

.github/workflows/publish.yml
```diff
@@ -33,14 +33,14 @@ jobs:
 
     - name: Bump react version patch
       run: |
-        grep -E "[0-9]+\.[0-9]+\.[0-9]+-beta\.[0-9]+" ./packages/react-icons/package.json | python3 -c """
+        grep -E "[0-9]+\.[0-9]+\.[0-9]+-rc\.[0-9]+" ./packages/react-icons/package.json | python3 -c """
```
spencer-nelson, line 36: Rather than replacing 'beta' with 'rc' in the grep, can you just do it as either beta or rc? Along similar lines, can you just make that bit conditionally match, so it doesn't have to change for a future update that is beta/rc/actual release?

.github/workflows/publish.yml
```diff
@@ -33,14 +33,14 @@ jobs:
...
         import sys
         import os
         current_version = sys.stdin.read().strip().split(':')[1]
         current_version = current_version.split('\"')[1]
-        major,minor,asset,beta = current_version.split('.')
-        asset = asset.replace('-beta','')
-        os.system(f'echo \"REACT_VERSION={major}.{minor}.{int(asset) + 1}-beta.{int(beta) + 1}\" >> \$GITHUB_ENV')
+        major,minor,asset,rc = current_version.split('.')
+        asset = asset.replace('-rc','')
+        os.system(f'echo \"REACT_VERSION={major}.{minor}.{int(asset) + 1}-rc.{int(rc) + 1}\" >> \$GITHUB_ENV')
```
spencer-nelson, line 43: This has me thinking we ought to find a solution that more easily allows beta/rc/release, but that can sit on the back burner if you don't immediately think of a better solution

spencer-nelson (changes requested):

spencer-nelson (approved):

PR #435, opened 2022-04-20
Fix Grep on macOS 12

GyroJoe:
Swaps the order of the include and exclude arguments to grep.

The version of grep on macOS 12 appears to apply the include/exclude arguments in order.

With `--exclude` first:
```
% grep --recursive   --ignore-case  --extended-regexp   --exclude="FluentIcon.swift" --include="*.swift" "\.[a-zA-Z0-9]+[0-9]{2}(Regular|Filled)" . | wc -l 
   22800
```

With `--exclude` last:
```
% grep --recursive   --ignore-case  --extended-regexp   --include="*.swift" --exclude="FluentIcon.swift"  "\.[a-zA-Z0-9]+[0-9]{2}(Regular|Filled)" . | wc -l
     902
```

In this change:
- Swapped the arguments produced in `Grep`
- Rebuilt the `run` binary
- Added a test `FluentIcon.swift` to the TestProject to catch behavior changes around `--exclude` in the future.


aaroncrespo:
@tomi-msft FYI this fixes some un-intended binary size increases we're seeing in our Xcode 13.3.1 (macOS 12) tests.

spencer-nelson (approved):

PR #436, opened 2022-04-20
Add dotnet-blazor package

vnbaaij:
The goal of this 'package' is to be able to create the FluentUI System Icons as optimized plain svg assets in a folder structure (assets name with spaces removed). The result is included as static assets in the official Microsoft (Fast) FluentUI Blazer component library. See [the FluentUI Blazore package](https://github.com/microsoft/fast-blazor) for reference.

No actual package is created when running the `npm build` command.


spencer-nelson:
This looks like it mostly boils down to running the same thing that happens in the svg-icons package with a slightly modified generate.js. Rather than creating a new folder in packages that doesn't actually publish anything, would it be better to add a way to direct generate.js to include the folders as needed, and then just use a clone/fork of the repository as part of the blazor pipeline(s)?

vnbaaij:
Yes, that is exactly what it is doing. I'm fine with any solution as long as we end up with an as frictionless way to generate and get the structure when new/updated icons have been released. But weary that with clone or fork things might end up being overlooked or overwritten.

vnbaaij:
Looking at this again, it is true that the crux of this is a variation on `generate.js`, but to get the needed end result, it also needs to do all the other steps as in the `svg-icons` package (clean, copy, rename, optimize and unfill). 
Not all the needed modules are there as dependencies in the `importer\package.json` file. Do we want to add them there? Or does it then make more sense to have this as a self-contained 'package'?

vnbaaij:
I have extended the `generate.js` file by adding a parameter that only needs to be set if you want the folder structure to be kept. I've also extended the `package.json` in the `svg-icons` folder with two extra scripts. One for calling `generate.js` with this new parameter set (`copywithdirs`) and one for building with the added first script (`buildforblazor`). I also changed the `rename` and `optimize` steps to make sure they handle the changed folder structure.

spencer-nelson (approved):

williamw2 (approved):

PR #438, opened 2022-04-28
Fluent icon font work

MLoughry:
This PR implements the work discussed previously to add an icon font implementation of `@fluentui/react-icons` as an alternative, which can be swapped in by using the `fluentIcontFont` import condition. It also adds a `webpack` plugin to subset the font files used in this case.

Change highlights:
- Upgraded `icon-font-generator` to its successor `fantasticon`, as specified in the former's NPM page.
  - (BREAKING?) The emitted JSON files now map string to codepoint as a number, rather than string to string of hexadecimal representation of the codepoint
- (BREAKING?) Fixed existing codepoint issues in the fonts, where the font generation tool would wrap values >0xFFFF back around due to overflow, or where certain glyphs would not render correctly because they were using control character codepoints. Now all codepoints fall in the "private use range" of Unicode between 0xE000 and 0xF8FF.
  - Additionally removed icons that no longer exist from the emitted codepoint JSON files.
- Added a new font file for "unsized" icons, following the same algorithm as used to create `packages/react-icons/src/icons` by using only 20px SVGs
- Added `.woff` and `.woff2` versions of the fonts.
- Added a script to generate a font implementation of the React icons under `react-icons/src/fonts`
- (BREAKING?) Added an export map to `@fluentui/react-icons`, restricting all imports to either the index or to specifically the SVG or font implementation. 
  - The index import points to the SVG implementation by default, but can swap to the font implementation using the `"fluentIcontFont"` condition.
- (INFRASTRUCTURE) Moved to using an npm workspace for the whole repository, rather than having separate `package-lock.json` files.
  - This necessitated using at least Node 16 for `npm>7.0.0`. As a result, I updated all pipeline definitions to use the latest stable version, Node 18, which I developed this change on.
  - This also allows for linking packages together within the repository.
- Created an (unpublished) test app (`packages/icon-app`) that renders all the React icons, showing the font and svg versions side-by-side. You can see a demonstration build of this here: https://owamaildev.blob.core.windows.net/owamail-branch/fluent-system-icons/index.html
- Added `packages/react-icons-font-subsetting-webpack-plugin` to allow for the subsetting of the font files based on which icons are used in the build.


.github/workflows/publish.yml
```diff
@@ -113,18 +113,30 @@ jobs:
         sed -i.bk -r "s/\"version\": \"[0-9]+\.[0-9]+\.[0-9]+(-beta\.[0-9]+)?(-rc\.[0-9]+)?\"/\"version\": \"$REACT_VERSION\"/g" packages/react-icons/package.json
         rm packages/react-icons/package.json.bk
 
+    # Needs to be "-E" instead of "-r" on macOS
+    - name: Replace version number in react-icons/package.json
```
spencer-nelson, line 117: Nitpick: This needs to be "...in react-icons-font-subsetting-webpack-plugin"

importer/generateFont.js
```diff
@@ -0,0 +1,107 @@
...
+const ICON_TYPE = argv.iconType;
+const CODEPOINTS_FILE = argv.codepoints;
+
+if (!SRC_PATH) {
+    throw new Error("SVG source folder not specified by --source");
+}
+if (!DEST_PATH) {
+    throw new Error("Output destination folder not specified by --dest");
+}
+if (!(ICON_TYPE === 'Filled' || ICON_TYPE === 'Regular' || ICON_TYPE === 'OneSize')) {
```
spencer-nelson, line 26: Rather than OneSize we should probably keep with the previous naming pattern.... I believe it was Resizable?

importer/generateFont.js
```diff
@@ -0,0 +1,107 @@
...
+    // Generate the font and associated assets
+    await fantasticon.generateFonts({
+        inputDir: stagingFolder,
+        outputDir: path.resolve(DEST_PATH),
+        name: `FluentSystemIcons-${ICON_TYPE}`,
+        fontTypes: [fantasticon.ASSET_TYPES.WOFF2, fantasticon.ASSET_TYPES.WOFF, fantasticon.ASSET_TYPES.TTF],
+        assetTypes: [fantasticon.ASSET_TYPES.CSS, fantasticon.ASSET_TYPES.HTML, fantasticon.ASSET_TYPES.JSON],
+        formatOptions: { json: { indent: 2 } },
+        codepoints: await getCodepoints(icons),
+        fontHeight: 500,
```
spencer-nelson, line 56: Is there something special about 500 here?

package.json
```diff
@@ -0,0 +1,15 @@
...
+  "workspaces": [
+    "packages/*",
+    "importer"
+  ],
+  "description": "![CI](https://github.com/microsoft/fluentui-system-icons/workflows/CI/badge.svg)",
+  "version": "1.0.0",
+  "main": "index.js",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/MLoughry/fluentui-system-icons.git"
```
spencer-nelson, line 13: Is this the right URL?

importer/generateFont.js
MLoughry: This was the value used previously in the call to `icon-font-generator`

packages/react-icons/src/utils/fonts/createFluentFontIcon.tsx
```diff
@@ -0,0 +1,99 @@
...
+    font-family: ${FONT_FAMILY_MAP[FontFile.Resizable]};
+    src: url(${JSON.stringify(fontOneSizeWoff2)}) format("woff2"),
+    url(${JSON.stringify(fontOneSizeWoff)}) format("woff"),
+    url(${JSON.stringify(fontOneSizeTtf)}) format("truetype");
+}
+`)
+
+const useRootStyles = makeStyles({
+    root: {
+        display: 'inline-block',
```
spmonahan, line 53: The SVG icon sets `display: 'inline'`. Should we use the same here?

MLoughry: I don't think so. It will still be treated as `inline` externally, but `inline-block` allows it to better control its own dimensions with the `line-height`

tomi-msft (approved):

PR #440, opened 2022-05-05
React icons script optimization

tomi-msft:
This PR is to optimize the `wrapIcon` script


packages/react-icons/convert.js
```diff
@@ -93,17 +93,17 @@ function processFolder(srcPath, destPath, resizable) {
...
     typescript: true,
     icon: true
   }
 
   var svgrOptsSizedIcons = {
     template: fileTemplate,
     expandProps: false, // HTML attributes/props for things like accessibility can be passed in, and will be expanded on the svg object at the start of the object
-    svgProps: { className: '{className}'}, // In order to provide styling, className will be used
-    replaceAttrValues: { '#212121': '{primaryFill}' }, // We are designating primaryFill as the primary color for filling. If not provided, it defaults to null.
+    svgProps: { className: '{undefined}'}, // In order to provide styling, className will be used
```
spencer-nelson, line 105: What is the overall effect of setting this undefined? Does it mean that people can no longer use classes to style/color the icons?

packages/react-icons/convert.js
```diff
@@ -134,14 +134,10 @@ function processFolder(srcPath, destPath, resizable) {
...
 `
-const ${destFilename}Icon = (iconProps: FluentIconsProps) => {
-  const { className, primaryFill } = iconProps;
-  return ${jsxCode};
-}
 
-export const ${destFilename} = /*#__PURE__*/wrapIcon(/*#__PURE__*/${destFilename}Icon({}), '${destFilename}');
-      `
 
+export const ${destFilename} = /*#__PURE__*/wrapIcon(/*#__PURE__*/${jsxCode}, '${destFilename}');
```
spencer-nelson, line 139: With the removal here of iconProps, does that impact the way that anything like the aria or other custom properties might be applied to the asset?

packages/react-icons/convert.js
```diff
@@ -155,7 +151,6 @@ export const ${destFilename} = /*#__PURE__*/wrapIcon(/*#__PURE__*/${destFilename
 
   for(const chunk of iconChunks) {
     chunk.unshift(`import wrapIcon from "../utils/wrapIcon";`)
-    chunk.unshift(`import { FluentIconsProps } from "../utils/FluentIconsProps.types";`)
```
spencer-nelson, line 158: If removing this, does the FluentIconsProps.types file also need to be removed, or at least the typedef?

packages/react-icons/convert.js
tomi-msft: `classname` and `fill` are still being set, they are just not being set using svgo. This allows for the svg to be directly passed into the `wrapIcon` method without having to explicitly define those values. They will then be defined in the `wrapIcon` method

packages/react-icons/convert.js
tomi-msft: No, those props will still be applied in the `wrapIcon` method. They were unnecessary here, and actually increased bundle size without needing to

packages/react-icons/convert.js
tomi-msft: The `FluentIconsProps.types` still needs to be defined, it just doesn't need to be used to define `className` and `primaryFill` props anymore, thus is unnecessary for the chunk files

spencer-nelson (approved):

PR #473, opened 2022-06-30
(chore): Move tslib and griffel/react to dependencies

tomi-msft:
This PR moves `tslib` and `@griffel/react` from `devDependencies` to `dependencies`. This is for projects with more stringent checks on deps being set correctly in a package.


spencer-nelson (approved):

packages/react-icons/package.json
```diff
@@ -44,10 +43,13 @@
...
-    "tslib": "^2.1.0",
     "typescript": "~4.1.0",
     "yargs": "^14.0.0"
   },
+  "dependencies": {
+    "@griffel/react": "^1.0.0",
+    "tslib": "^2.1.0"
+  },
   "peerDependencies": {
     "@griffel/react": "^1.0.0",
```
layershifter, line 54: Can you please clarify why `@griffel/react` is still in `peerDependencies`? Is it intentional?

PR #487, opened 2022-08-30
Add codepoint property to fluent icon component

danivarelas:
Export the codepoint that is matched against the icon

### DIFF
--- a/packages/react-icons/src/utils/fonts/createFluentFontIcon.tsx
+++ b/packages/react-icons/src/utils/fonts/createFluentFontIcon.tsx
@@ -95,5 +95,6 @@ export function createFluentFontIcon(displayName: string, codepoint: string, fon
         return <i {...state}>{codepoint}</i>
     }
     Component.displayName = displayName;
+    Component.codepoint = codepoint;
     return Component;
 }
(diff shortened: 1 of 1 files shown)


MLoughry (approved):

packages/react-icons/src/utils/fonts/createFluentFontIcon.tsx
```diff
@@ -95,5 +95,6 @@ export function createFluentFontIcon(displayName: string, codepoint: string, fon
         return <i {...state}>{codepoint}</i>
     }
     Component.displayName = displayName;
+    Component.codepoint = codepoint;
```
MLoughry, line 98: To fix the build, I think you need to change the type on line 73 (and maybe also 72) to
```ts
React.FC<FluentIconsProps<React.HTMLAttributes<HTMLElement>>> & { codepoint: string}
```

danivarelas: @MLoughry pinging you about this update, thanks

MLoughry (approved):

packages/react-icons/src/utils/fonts/createFluentFontIcon.tsx
MLoughry: I aproved, but you'll need a review with someone with write access (like @tomi-msft)

tomi-msft (approved):

PR #505, opened 2022-10-24
Forward ref

### LINKED ISSUE #504: wrapIcon not forwarding ref
Can we add the ability to forward ref on [`wrapIcon`](https://github.com/microsoft/fluentui-system-icons/blob/main/packages/react-icons/src/utils/wrapIcon.tsx)?

fluent ui react v9 tooltip/popover requires trigger to forward ref. So when our icon in @fluentui/react-icon is used as v9 tooltip/popover's trigger, the tooltip/popover is positioned wrongly :(
https://codesandbox.io/s/charming-yalow-1dpsug?file=/example.tsx

tomi-msft:
This PR is to add ref forwarding to the build script for the `react-icons` package. 
Fixes (https://github.com/microsoft/fluentui/issues/24544).
Also fixes #504 issue


spencer-nelson (approved):

packages/react-icons/convert.js
```diff
@@ -135,10 +135,10 @@ function processFolder(srcPath, destPath, resizable) {
       var jsCode = 
 `
 
-const ${destFilename}Icon = (props: FluentIconsProps) => {
+const ${destFilename}Icon = React.forwardRef((props: FluentIconsProps, ref: React.LegacyRef<SVGSVGElement>) => {
```
behowell, line 138: This should be `React.Ref` instead of `React.LegacyRef`. Also, why is this using `SVGSVGElement` here, but just `SVGElement` in wrapIcon?

packages/react-icons/convert.js
```diff
@@ -93,7 +93,7 @@ function processFolder(srcPath, destPath, resizable) {
   var svgrOpts = {
     template: fileTemplate,
     expandProps: 'start', // HTML attributes/props for things like accessibility can be passed in, and will be expanded on the svg object at the start of the object
-    svgProps: { className: '{className}'}, // In order to provide styling, className will be used
+    svgProps: { ref: '{ref}'}, // In order to provide styling, className will be used
```
jurokapsiar, line 96: the comment does not seem to be valid anymore

packages/react-icons/convert.js
```diff
@@ -135,10 +135,10 @@ function processFolder(srcPath, destPath, resizable) {
       var jsCode = 
 `
 
-const ${destFilename}Icon = (props: FluentIconsProps) => {
-  const { fill: primaryFill = 'currentColor', className } = props;
```
jurokapsiar, line 139: wasn't className used before? I can see it being removed from the destructuring only.

packages/react-icons/src/utils/wrapIcon.tsx
```diff
@@ -2,11 +2,15 @@ import * as React from "react";
 import { FluentIconsProps } from "./FluentIconsProps.types";
 import { useIconState } from "./useIconState";
 
-const wrapIcon = (Icon: (iconProps: FluentIconsProps) => JSX.Element, displayName?: string) => {
-    const WrappedIcon = (props: FluentIconsProps) => { 
+type IconWithRefProps = 
+    React.ForwardRefExoticComponent<FluentIconsProps & 
+    React.RefAttributes<SVGSVGElement>>;
+
+const wrapIcon = (Icon: IconWithRefProps, displayName?: string) => {
```
jurokapsiar, line 9: please also test if this works for bundled icons - as far as I remember we render two icons next to each other - not sure which one should obtain the ref.

tomi-msft:
fixed by #583

PR #525, opened 2022-11-24
fix(react-icons): allow React 18 in peerDependencies

### LINKED ISSUE #517: Upgrade react 18 peer dependency
@fluentui/react-components and @griffle/react have already updated theirs react peer dependencies. Can we do the same here? https://github.com/microsoft/fluentui-system-icons/blob/main/packages/react-icons/package.json#L55 

fluentui/react-componnets
https://github.com/microsoft/fluentui/issues/22241#issuecomment-1278137151

Griffle: 
https://github.com/microsoft/griffel/issues/262

layershifter:
Fixes #517.
See title 👆

### DIFF
--- a/packages/react-icons/package.json
+++ b/packages/react-icons/package.json
@@ -51,8 +51,7 @@
     "tslib": "^2.1.0"
   },
   "peerDependencies": {
-    "@griffel/react": "^1.0.0",
-    "react": ">=16.8.0 <=18.0.0"
+    "react": ">=16.8.0 <19.0.0"
   },
   "files": [
     "lib/",
@@ -83,3 +82,4 @@
     }
   }
 }
+


packages/react-icons/package.json
```diff
@@ -52,7 +52,7 @@
   },
   "peerDependencies": {
     "@griffel/react": "^1.0.0",
-    "react": ">=16.8.0 <=18.0.0"
+    "react": ">=16.8.0 <19.0.0"
```
layershifter, line 55: Allows React 18, the same as in `@fluentui/react-components`, https://github.com/microsoft/fluentui/blob/8fcefaeacac2e2cc2cd7ca5b76c1ca5e7e547682/packages/react-components/react-components/package.json#L80.

packages/react-icons/package.json
```diff
@@ -51,8 +51,7 @@
     "tslib": "^2.1.0"
   },
   "peerDependencies": {
-    "@griffel/react": "^1.0.0",
```
layershifter, line 54: `@griffel/react` is gone as it's also in regular dependencies (see line 50).

layershifter:
@tomi-msft can you please take a look?

tomi-msft (approved):

richardkooiman:
@tomi-msft Thanks for the merge, encountered the same issue so happy with this fix. Any idea when we can expect a new release which includes this fix?

PR #529, opened 2022-12-15
Added package for react-native SVG icons

ValentinaKozlova:
Added package for generation of react-native SVG icons.
1st commit contains copy of react-icons code because majority of it can be reused.
For better readability in other commits I added react-native-specific code.
This PR is for getting early feedback.
Maybe it would be good to put duplicated code in sort of `react-common` folder.

**Issues with current package:**
1. Fails while running `npm run build:cjs`
2. Icon is generated with size in `rm` which is not suitable for RN

**Things to do:**
1. Fix issues
2. Add font icons support
3. Move duplicated code to a separate folder

Verification:
Icon generated code looks like this:
```jsx
import * as React from "react";
import { FluentReactNativeIconsProps } from "../utils/FluentReactNativeIconsProps.types";
import wrapIcon from "../utils/wrapIcon";
import { Path, Svg, Rect, G } from 'react-native-svg';


const AccessTimeFilledIcon = (props: FluentReactNativeIconsProps) => {
  const { fill: primaryFill = 'currentColor', style } = props;
  return <Svg {...props} width="1em" height="1em" viewBox="0 0 20 20" fill="none" xmlns="http://www.w3.org/2000/svg" style={style}><Path d="M...7Z" fill={primaryFill} /></Svg>;
}
export const AccessTimeFilled = /*#__PURE__*/wrapIcon(/*#__PURE__*/AccessTimeFilledIcon, 'AccessTimeFilled');
 ```

In RN tester:
image

### CODE REVIEW

File: packages/react-native-icons/convert.js [hunk 1]
```diff
@@ -132,12 +132,16 @@ function processFolder(srcPath, destPath, resizable) {
       var iconContent = fs.readFileSync(srcFile, { encoding: "utf8" })
       
       var jsxCode = resizable ? svgr.default.sync(iconContent, svgrOpts, { filePath: file }) : svgr.default.sync(iconContent, svgrOptsSizedIcons, { filePath: file })
+      var rnRegex = new RegExp('(<(|\/))(svg|path|rect|g)', 'g')
```
ValentinaKozlova, line 135: As majority of our icons contains only those tags I decided to use regex to change casing

packages/react-native-icons/convert.js
```diff
@@ -69,9 +69,9 @@ function processFiles(src, dest) {
   // Finally add the interface definition and then write out the index.
   indexContents.push('export { FluentReactNativeIconsProps } from \'./utils/FluentReactNativeIconsProps.types\'');
   indexContents.push('export { default as wrapIcon } from \'./utils/wrapIcon\'');
-  indexContents.push('export { default as bundleIcon } from \'./utils/bundleIcon\'');
+  //indexContents.push('export { default as bundleIcon } from \'./utils/bundleIcon\'');
```
rurikoaraki, line 72: Should these commented lines be removed?

rurikoaraki:
Should a "Build React Native library" CI step be added?

rurikoaraki (approved):

ValentinaKozlova:
> Should a "Build React Native library" CI step be added?

Yes, but I thought to add it separately. Next steps can be to move duplicate code to a separate folder, check the generated icons and then build it for RN

packages/react-native-icons/convert.js
ValentinaKozlova: Not for now. Maybe I'll need it later. If not, I'll remove it

PR #537, opened 2023-01-12
Title prop fix

tomi-msft:
This PR adds a title prop to the built icons in the react-icons package


spencer-nelson (approved):

tomi-msft:
Adding a `title` prop to the icons will be a net negative for accessibility as it will require all icons to conditionally render the <title> tag to properly support localization, plus not all icons should render a title. Such a feature is likely to be a performance regression.

Closing this PR with this consideration in mind.

PR #550, opened 2023-02-24
Additional bidirectional arrows

2metres:


2metres:

spencer-nelson:
Thanks for the contribution. We've grabbed the SVGs and will work with our designers to get them reviewed and added to the Figma library with the next sprint or two.

Issue #553, opened 2023-03-06
Feature: Support auto-mirroring React icons 


behowell:
Feature request: https://github.com/microsoft/fluentui/issues/25624

```[tasklist]
### Tasks
- [ ] Add support for applying mirroring styles (`transform: 'scaleX(-1)'`) when generating the React icon
- [ ] Fill out the list of icons that should flip: [rtl.txt](https://github.com/microsoft/fluentui-system-icons/blob/main/importer/rtl.txt) (might need to create a new list -- check how this one is currently used)
- [ ] (Maybe) Add a prop to control the flipping behavior (opt-out of auto-flipping)
- [ ] Add support for conditionally rendering a different SVG based on RTL (e.g. `BookQuestionMark24Regular` => `BookQuestionMarkRtl24Regular`)
- [ ] (Maybe) Export a function to create icons that swap SVGs based on RTL (similar to `bundleIcon`)
```

PR #554, opened 2023-03-09
React icons rtl support

tomi-msft:
This PR is the first step in adding auto-mirroring to react-icons. It use `griffel/react` `useTextDirection` hook to detect `rtl` and applies styling based on that.

Part of #553


packages/react-icons/src/utils/useIconState.tsx
```diff
@@ -1,5 +1,6 @@
 import { FluentIconsProps } from "./FluentIconsProps.types";
 import { makeStyles, mergeClasses } from "@griffel/react";
+import { useTextDirection } from '@griffel/react/TextDirectionContext'
```
behowell, line 3: We'll need to have this function exported from griffel before we can actually use it.

packages/react-icons/src/utils/useIconState.tsx
```diff
@@ -22,7 +27,7 @@ export const useIconState = <TBaseAttributes extends (React.SVGAttributes<SVGEle
   
     const styles = useRootStyles();
   
-    state.className = mergeClasses(styles.root, state.className);
+    state.className = mergeClasses(styles.root, dir === 'rtl' && styles.rtl, state.className);
```
behowell, line 30: We need a way to only apply this mirroring behavior to icons that need it. One option is a prop like `autoMirror` that defaults to false. We could then change the default to true for specific icons from some list (like rtl.txt) when generating the React components for the icons.

behowell (changes requested):
Need to resolve the griffel deep import, and add a way to only apply this to certain icons, before merging. Thanks!

tomi-msft:
Closing in favor of a cleaner commit history branch

Issue #561, opened 2023-03-19
Package size of `react-icons` is optimizable


kkocdko:
Now it produce the code like this:

```ts
const AccessibilityCheckmarkFilledIcon = props => {
  const {
    fill: primaryFill = 'currentColor',
    className
  } = props;
  return React.createElement("svg", Object.assign({}, props, {
    width: "1em",
    height: "1em",
    viewBox: "0 0 20 20",
    xmlns: "http://www.w3.org/2000/svg",
    className: className
  }), React.createElement("path", {
    d: "M10 6a2 2 0 1 0 0...",
    fill: primaryFill
  }));
};
// And many many repeats...
```

I think we can reduce the duplicate code by using a helper function.

```ts
export const AccessibilityCheckmarkFilledIcon = createFluentIcon('AccessibilityCheckmarkFilledIcon', 20, ['M10 6a2 2 0 1 0 0...']);
// And more
```

The `lib-cjs` directory also cause double size.

I'll start to working for this.

- [x] Reduce the duplicate code of `react-icons` by using a helper function.
- [ ] Consider to drop commonjs support?

PR #562, opened 2023-03-19
Reduce react-icons package size

kkocdko:
Define icons by the new `createFluentIcon` helper function, to reduce the npm package size and bundle size.

The `./lib` dir: 21.3 MiB -> 9.85 MiB.

Issue https://github.com/microsoft/fluentui-system-icons/issues/561

### DIFF
--- a/packages/react-icons/convert.js
+++ b/packages/react-icons/convert.js
@@ -1,10 +1,8 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
-const svgr = require("@svgr/core");
 const fs = require("fs");
 const path = require("path");
-const process = require("process");
 const argv = require("yargs").boolean("selector").default("selector", false).argv;
 const _ = require("lodash");
 
@@ -86,26 +84,6 @@ function processFiles(src, dest) {
  */
 function processFolder(srcPath, destPath, resizable) {
   var files = fs.readdirSync(srcPath)
-
-  // These options will be passed to svgr/core
-  // See https://react-svgr.com/docs/options/ for more info
-  var svgrOpts = {
-    template: fileTemplate,
-    expandProps: 'start', // HTML attributes/props for things like accessibility can be passed in, and will be expanded on the svg object at the start of the object
-    svgProps: { className: '{className}'}, // In order to provide styling, className will be used
-    replaceAttrValues: { '#212121': '{primaryFill}' }, // We are designating primaryFill as the primary color for filling. If not provided, it defaults to null.
-    typescript: true,
-    icon: true,
-  }
-
-  var svgrOptsSizedIcons = {
-    template: fileTemplate,
-    expandProps: 'start', // HTML attributes/props for things like accessibility can be passed in, and will be expanded on the svg object at the start of the object
-    svgProps: { className: '{className}'}, // In order to provide styling, className will be used
-    replaceAttrValues: { '#212121': '{primaryFill}' }, // We are designating primaryFill as the primary color for filling. If not provided, it defaults to null.
-    typescript: true,
-  }
-
   /** @type string[] */
   const iconExports = [];
   files.forEach(function (file, index) {
@@ -129,17 +107,10 @@ function processFolder(srcPath, destPath, resizable) {
       destFilename = destFilename.replace(destFilename.substring(0, 1), destFilename.substring(0, 1).toUpperCase()) // capitalize the first letter
 
       var iconContent = fs.readFileSync(srcFile, { encoding: "utf8" })
-      
-      var jsxCode = resizable ? svgr.default.sync(iconContent, svgrOpts, { filePath: file }) : svgr.default.sync(iconContent, svgrOptsSizedIcons, { filePath: file })
-      var jsCode = 
-`
-
-const ${destFilename}Icon = (props: FluentIconsProps) => {
-  const { fill: primaryFill = 'currentColor', className } = props;
-  return ${jsxCode};
-}
-export const ${destFilename} = /*#__PURE__*/wrapIcon(/*#__PURE__*/${destFilename}Icon, '${destFilename}');
-      `
+      const getAttr = (key) => [...iconContent.matchAll(`(?<= ${key}=)".+?"`)].map((v) => v[0]);
+      const width = resizable ? '"1em"' : getAttr("width")[0];
+      const paths = getAttr("d").join(',');
+      var jsCode = `export const ${destFilename} = /*#__PURE__*/createFluentIcon('${destFilename}', ${width}, [${paths}]);`
       iconExports.push(jsCode);
     }
   });
@@ -148,13 +119,12 @@ export const ${destFilename} = /*#__PURE__*/wrapIcon(/*#__PURE__*/${destFilename
   /** @type string[][] */
   const iconChunks = [];
   while(iconExports.length > 0) {
-    iconChunks.push(iconExports.splice(0, 500));
+    iconChunks.push(iconExports.splice(0, 1000));
   }
 
   for(const chunk of iconChunks) {
-    chunk.unshift(`import wrapIcon from "../utils/wrapIcon";`)
-    chunk.unshift(`import { FluentIconsProps } from "../utils/FluentIconsProps.types";`)
-    chunk.unshift(`import * as React from "react";`)
+    chunk.unshift(`import { createFluentIcon } from "../utils/createFluentIcon";`);
+    chunk.unshift(`import type { FluentIcon } from "../utils/createFluentIcon";`);
   }
 
   /** @type string[] */
--- a/packages/react-icons/convert.js
+++ b/packages/react-icons/convert.js
@@ -110,7 +110,7 @@ function processFolder(srcPath, destPath, resizable) {
       const getAttr = (key) => [...iconContent.matchAll(`(?<= ${key}=)".+?"`)].map((v) => v[0]);
       const width = resizable ? '"1em"' : getAttr("width")[0];
       const paths = getAttr("d").join(',');
-      var jsCode = `export const ${destFilename} = /*#__PURE__*/createFluentIcon('${destFilename}', ${width}, [${paths}]);`
+      var jsCode = `export const ${destFilename} = (/*#__PURE__*/createFluentIcon('${destFilename}', ${width}, [${paths}]));`
       iconExports.push(jsCode);
     }
   });
--- a/packages/react-icons/src/utils/createFluentIcon.ts
+++ b/packages/react-icons/src/utils/createFluentIcon.ts
@@ -3,24 +3,24 @@ import { FluentIconsProps } from "./FluentIconsProps.types";
 import { useIconState } from "./useIconState";
 
 export type FluentIcon = {
-    (props: FluentIconsProps<React.SVGAttributes<SVGElement>>): JSX.Element;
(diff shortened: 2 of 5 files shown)


packages/react-icons/src/utils/createFluentIcon.tsx
```diff
@@ -0,0 +1,31 @@
+import * as React from "react";
+import { FluentIconsProps } from "./FluentIconsProps.types";
+import wrapIcon from "./wrapIcon";
+
+export type FluentIcon = {
+    (props: FluentIconsProps<React.SVGAttributes<SVGElement>>): JSX.Element;
+    displayName: string | undefined;
+}
+
+export const createFluentIcon = (displayName: string, width: string, paths: string[]): FluentIcon => {
```
kkocdko, line 10: Function name is similar to `./src/utils/fonts/createFluentFontIcon` helper function.

packages/react-icons/src/utils/createFluentIcon.tsx
```diff
@@ -0,0 +1,31 @@
+import * as React from "react";
+import { FluentIconsProps } from "./FluentIconsProps.types";
+import wrapIcon from "./wrapIcon";
+
+export type FluentIcon = {
```
kkocdko, line 5: Also reduce the size of `chunk-x.d.ts`. If not, every icon will produce very long and same type define. Prefer to use `type` instead of `interface` here because it's more transparent and avoid breaking changes in some extreme case.

packages/react-icons/convert.js
```diff
@@ -1,10 +1,8 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
-const svgr = require("@svgr/core");
```
kkocdko, line 4: Remove `svgr` because now we only want to extract data but never transform the SVG.

packages/react-icons/convert.js
```diff
@@ -149,13 +121,11 @@ export const ${destFilename} = /*#__PURE__*/wrapIcon(/*#__PURE__*/${destFilename
   /** @type string[][] */
   const iconChunks = [];
   while(iconExports.length > 0) {
-    iconChunks.push(iconExports.splice(0, 500));
```
kkocdko, line 152: The file size is much smaller now, put more icon into one file will be a little faster.

kkocdko:

kkocdko:
Seems conflict with https://github.com/microsoft/fluentui-system-icons/pull/505 . Maybe this improvement should be taken after that pr was merged?

I hope the progress be faster.

layershifter:
@kkocdko we have an issue on Fluent UI side related to this problem, https://github.com/microsoft/fluentui/issues/25989. The idea with modifying glyphs is pretty nice. However, it requires more checks to ensure that nothing else will regress.

Can you please split this PR to two separate ones? I.e. make types changes and code changes separate. With that we could progress faster.

packages/react-icons/src/utils/createFluentIcon.tsx
layershifter: That's great improvement, it's what I also mentioned when was checking this area.

However, I think that it could be improved even more. In this PR we have:

```ts
export declare const Accessibility16Regular: import("../utils/createFluentIcon").FluentIcon;
export declare const Accessibility20Filled: import("../utils/createFluentIcon").FluentIcon;
export declare const Accessibility20Regular: import("../utils/createFluentIcon").FluentIcon;
````

But we can do following:

```ts
export declare const AccessibilityRegular: FluentIcon;
export declare const AccessibilityCheckmarkFilled: FluentIcon;
export declare const AccessibilityCheckmarkRegular: FluentIcon;
```

We need to do following changes to achieve that:

```diff
  for(const chunk of iconChunks) {
    chunk.unshift(`import { createFluentIcon } from "../utils/createFluentIcon";`);
+   chunk.unshift(`import type { FluentIcon } from "../utils/createFluentIcon";`);
  }
```

```diff
-      var jsCode = `export const ${destFilename} = /*#__PURE__*/createFluentIcon('${destFilename}', ${width}, [${paths}]);`
+      var jsCode = `export const ${destFilename}: FluentIcon = /*#__PURE__*/createFluentIcon('${destFilename}', ${width}, [${paths}]);`
```

layershifter:
BTW, while I can agree about dist size improvements, improvements in bundle size are not so great 🐱 

```
┌────────────────────────────────┬───────────────┬───────────┐
│ Fixture                        │ Minified size │ GZIP size │
├────────────────────────────────┼───────────────┼───────────┤
│ react-icons (new): single icon │ 5.844 kB      │ 2.659 kB  │
├────────────────────────────────┼───────────────┼───────────┤
│ react-icons (new): 5 icons     │ 7.765 kB      │ 3.041 kB  │
├────────────────────────────────┼───────────────┼───────────┤
│ react-icons (old): single icon │ 5.726 kB      │ 2.603 kB  │
├────────────────────────────────┼───────────────┼───────────┤
│ react-icons (old): 5 icons     │ 8.541 kB      │ 3.028 kB  │
└────────────────────────────────┴───────────────┴───────────┘
```

kkocdko:
@layershifter I also tried your suggesstion:

```diff
for(const chunk of iconChunks) {
    chunk.unshift(`import { createFluentIcon } from "../utils/createFluentIcon";`);
+   chunk.unshift(`import type { FluentIcon } from "../utils/createFluentIcon";`);
}

-      var jsCode = `export const ${destFilename} = /*#__PURE__*/createFluentIcon('${destFilename}', ${width}, [${paths}]);`
+      var jsCode = `export const ${destFilename}: FluentIcon = /*#__PURE__*/createFluentIcon('${destFilename}', ${width}, [${paths}]);`
```

But what I shocked is, if I import the `type FluentIcon` and write `export const xxx: FluentIcon`, the `/*#__PURE__*/` mark will gone in bundle result! I don't know why and how to fix this.

packages/react-icons/tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     "outDir": "lib",
-    "target": "ES6",
+    "target": "esnext",
     "module": "esnext",
     "lib": ["es6", "dom"],
```
kkocdko, line 6: Reduce some size. Because we already use `"module": "esnext"` before, so no breaking changes.

packages/react-icons/src/utils/createFluentIcon.ts
```diff
@@ -0,0 +1,31 @@
...
+                    d,
+                    fill: props.fill, // We are designating primaryFill as the primary color for filling. If not provided, it defaults to null
+                    // key: i // The key for static children is needless 
+                })
+            )
+        );
+    }
+    Icon.displayName = displayName;
+    return Icon;
+}
```
sopranopillow, line 31: You could maybe simplify it into this?
```suggestion
export const createFluentIcon = (
  displayName: string,
  width: string,
  paths: string[]
): FluentIcon => {
  const viewBoxWidth = width === "1em" ? "20" : width;
  const Icon = (props: FluentIconsProps) => {
    const state = useIconState({
      ...props,
      width,
      height: width,
      viewBox: `0 0 ${viewBoxWidth} ${viewBoxWidth}`,
      xmlns: "http://www.w3.org/2000/svg",
    });

    return React.createElement(
      "svg",
      state,
      ...paths.map((d) =>
        React.createElement("path", {
          d,
          fill: state.fill, // We are designating primaryFill as the primary color for filling. If not provided, it defaults to null
        })
      )
    );
  };
  Icon.displayName = displayName;
  return Icon;
};
```

kkocdko: What's the key-point, you prefer `const state = xxx`? If so, this:

```js
const state = {
  ...useIconState(props),
  width,
  height: width,
  viewBox: `0 0 ${viewBoxWidth} ${viewBoxWidth}`,
  xmlns: "http://www.w3.org/2000/svg",
};
```

May be better. `useIconState` seems used to add accessibility attrs, the props like width and viewBox is only for svg element and does not need to be passed into `useIconState`.

sopranopillow: Mainly for readability and to not modify the props directly. You can change it to the snippet you added above, both work 🙂

kkocdko: Solved.

layershifter:
> But what I shocked is, if I import the `type FluentIcon` and write `export const xxx: FluentIcon`, the `/*#__PURE__*/` mark will gone in bundle result! I don't know why and how to fix this. Now fixed and your suggesstion applied.

Oh my gosh, this TS bug strikes again, it's fixed in newer versions of TS 🙈

@kkocdko you need to wrap it with parens:

```
var jsCode = `export const ${destFilename}: FluentIcon = (/*#__PURE__*/createFluentIcon('${destFilename}', ${width}, [${paths}]));`
```

Here is a [repro](https://www.typescriptlang.org/play?ts=4.5.5#code/C4TwDgpgBAYgNgVwgO2ASQMYHtlQLxQCGyIAUKQGYLIbACWOUGAThIcBPEqpjgBQA6IYWYBzAM4AuIiQCUUAN4BfctmTjgUAILSuKdGvxQA9ACoAxAH1LABQCqAJQCi108ZZsOenmr6yA3KRqGlAAQrqI+ry4BHxmVraOLpZuHuyckT78sgFAA).

layershifter:
> Maybe hard to do... Create a type FluentIcon in pr 1 and use it in pr 2 will cause build failed, looks unpleasant. And I think this pr with < 40 lines of code should not be split more.

Well, I personally looking for a bigger refactor 🐱

As the chain is strange now: `createFluentIcon()` => `wrapIcon()`=> `useIconState()`. For me, it looks like too much for an icon 🤔 However, it has nothing to do with your PR.

I don't have other comments to this PR, @kkocdko great job ⭐ If you would like to contribute more, let's continue on https://github.com/microsoft/fluentui/issues/25989.

sopranopillow (approved):

packages/react-icons/convert.js
```diff
@@ -129,17 +107,10 @@ function processFolder(srcPath, destPath, resizable) {
...
-const ${destFilename}Icon = (props: FluentIconsProps) => {
-  const { fill: primaryFill = 'currentColor', className } = props;
-  return ${jsxCode};
-}
-export const ${destFilename} = /*#__PURE__*/wrapIcon(/*#__PURE__*/${destFilename}Icon, '${destFilename}');
-      `
+      const getAttr = (key) => [...iconContent.matchAll(`(?<= ${key}=)".+?"`)].map((v) => v[0]);
+      const width = resizable ? '"1em"' : getAttr("width")[0];
+      const paths = getAttr("d").join(',');
+      var jsCode = `export const ${destFilename} = /*#__PURE__*/createFluentIcon('${destFilename}', ${width}, [${paths}]);`
```
layershifter, line 113:
```suggestion
      var jsCode = `export const ${destFilename} = (/*#__PURE__*/createFluentIcon('${destFilename}', ${width}, [${paths}]));`
```

layershifter: Let's apply this change before merging to preserve `PURE` annotations, https://github.com/microsoft/fluentui-system-icons/pull/562#issuecomment-1494039774.

tomi-msft (approved):

tomi-msft:
Thanks for your contribution! This was extremely helpful

PR #564, opened 2023-03-24
Update ios/README.md

zizicici:
Previous path is not working in Xcode pre-action

### DIFF
--- a/ios/README.md
+++ b/ios/README.md
@@ -55,7 +55,7 @@ At build/release time you can run the following script to ensure all unused asse
 CocoaPods
 
 ```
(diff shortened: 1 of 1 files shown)


nickromano:
Hey @zizicici, I actually wouldn't recommend running this in an Xcode build phase. Otherwise, you'll hit runtime failures in the app when the icons are missing. I'd recommend running this script on your CI or locally before you archive and build for release to strip out the unused assets (where we won't have access to ${PODS_ROOT}.

zizicici:
Hi @nickromano, Thanks for your reply.

Integration with FluentIcons Pods cost about 10 minutes in Xcode build phase.

What is the best practice to reduce the compile time?

PR #578, opened 2023-05-02
Move react-native-svg from dependencies to peerDependencies

verluci:
React-native-svg is notorious for breaking things when there's multiple versions of the package installed. It's also very frequently required for packages. Listing it as a peerDependency prevents these issues. Other libraries from the Fluentui stack follow this approach as well, for example https://github.com/microsoft/fluentui-react-native/blob/main/packages/components/Icon/package.json


verluci:

verluci:
Other libraries from the Fluentui stack follow this approach as well, for example https://github.com/microsoft/fluentui-react-native/blob/main/packages/components/Icon/package.json

verluci:
Would you mind taking a look at this pr, @nickromano or @willhou?

nickromano:
Hey @warren-ms can you please take a look?

tomi-msft (approved):

PR #596, opened 2023-07-01
Convert XML Fluent Icons to ImageVector Compose Icons

Younes-Charfaoui:
## What has been done 🔥
This pull request (PR) aims to enhance the icons set provided by fluent system icons converting existing XML icons to Jetpack Compose ImageVector icons. By migrating to Jetpack Compose, we can leverage the declarative nature of Compose UI and take advantage of its powerful features for rendering scalable icons. This conversion will result in improved performance, flexibility, and maintainability of the icons in the project.

## How it was done ✅
I have created a parser that parses the XML icons and converts them to `fluentIcons` using `ImageVector` compose's nature using `materialPath`.

## Changes Made 💡
- Converted existing XML icons to Jetpack Compose ImageVector format.
- Added necessary code to accommodate the changes introduced by the new icon format (especially the different sizes of the icons)

## Testing Done ✔
- Verified that the icons maintain their original appearance and fidelity after the conversion process.

I think we can also make some screenshot tests for each icon, but that would take a long time to run the tests.

## Missing Stuff 😅
Three icons are still remaining since they don't share the same structure as the other icons (they contain extra `clip-path`); I will try to convert those manually. Here are the icons:
- [ ] ic_fluent_my_location_12_filled.xml
- [ ] ic_fluent_my_location_12_regular.xml
- [ ] ic_fluent_mention_12_filled.xml


Younes-Charfaoui:

Younes-Charfaoui:
Forget to add the variation of the Filled and Regular, at the moment of generating the files; since they had the same names, there were some overwrites.

Issue #603, opened 2023-07-11
Metadata for icon sets


williamch-ms:
Adding metadata.json files to each icon folder within the assets directory.

PR #604, opened 2023-07-12
Metadata for A through B, removed some LTR/RTL folders, and merged in…

williamch-ms:
Metadata for A through B, removed some LTR/RTL folders, and merged Book and Book Default.
Issue: https://github.com/microsoft/fluentui-system-icons/issues/603
Folders removed:

Todo:
~~Re-export Arrow Export.~~

- Arrow Export LTR   => Arrow Export
- Arrow Export RTL   
- Arrow Repo Temp LTR
- Arrow Repo Temp RTL
- Arrow Undo Temp RTL
- Arrow Undo Temp LTR
- Book Default (merged into Book)


behowell (changes requested):
I don't think we can remove or rename the existing LTR/RTL icons, such as "Arrow Export LTR"/"Arrow Export RTL", as that will break anyone who's currently using them. We could remove them in a future major version of the icon package, but not at this time.

Instead, I think it will be necessary to keep the existing LTR and RTL variants of each icon, _without_ ~~`"directionType": "mirror"`~~.

Then _copy_ the LTR version to one without a suffix _with_ `"directionType": "mirror"`. (E.g. copy "Arrow Export LTR" to Arrow Export", with mirroring).

spencer-nelson:
The purpose of this work is explicitly to remove the RTL/LTR and will be a major change anyway. For the most part, the major versions to these libraries have already been thrown around enough with React having resizable icons and no other platform following up, iOS having language directionality support (which variance highly contributed to this particular need to remove files) and fonts being tied/adopted for some platforms but not others. Hence, should we just consider this directionality to be a breaking change anyway and deal with the tech debt of removal now, rather than continuing to punt it down the road?

behowell:
@spencer-nelson It would be ideal if we could add the auto-mirroring support without a major version bump, and then later delete the redundant icons with a major version. 

From an engineering perspective, partners picking up this change will need time to update their code to no longer refer to the LTR/RTL icons. By adding the new mirroring functionality _separately_ before removing the existing icons, we give them a chance to do that work over time, rather than requiring it to be done immediately.

Also, do you know if all consumers are able to auto-mirror the icons based on the metadata? There might still be a longer-term need for the separate RTL/LTR variants in places that don't (yet) support mirroring.

williamch-ms:
Closing, addressed feedback in https://github.com/microsoft/fluentui-system-icons/pull/618.

PR #625, opened 2023-08-09
Icons autoflip logic

tomi-msft:
This PR is to add styling to allow automirroring for specific icons marked by design to be able to be flipped in rtl. There will be a follow up PR to add switching to the rtl equivalent icon for the icons that require this in a later PR


packages/react-icons/src/contexts/IconDirectionContext.ts
```diff
@@ -10,4 +10,4 @@ const IconDirectionContextDefaultValue: IconDirectionContextValue = {};
 
 export const IconDirectionContextProvider = IconDirectionContext.Provider;
 
-export const useIconContext = () => React.useContext(IconDirectionContext) ?? IconDirectionContextDefaultValue;
\ No newline at end of file
+export const useIconContext = () => React.useContext(IconDirectionContext) ? IconDirectionContextDefaultValue : undefined;
```
MLoughry, line 13: I don't understand this change. This is saying that if the context has a value, use the default value instead; otherwise, use `undefiend`.

packages/react-icons/src/contexts/IconDirectionContext.ts
```diff
@@ -10,4 +10,4 @@ const IconDirectionContextDefaultValue: IconDirectionContextValue = {};
 
 export const IconDirectionContextProvider = IconDirectionContext.Provider;
 
-export const useIconContext = () => React.useContext(IconDirectionContext) ?? IconDirectionContextDefaultValue;
\ No newline at end of file
+export const useIconContext = () => React.useContext(IconDirectionContext) ? React.useContext(IconDirectionContext) : IconDirectionContextDefaultValue
```
MLoughry, line 13: I'm not sure I understand why you're getting rid of the existing nullish coalescing operator `??`. It's functionally the same

tomi-msft: Ahh yes this is because of an issue where the nullish coalescing operator causes an error with some webpack configurations, so I just write it out

packages/react-icons/package.json
```diff
@@ -16,16 +16,18 @@
...
     "generate:font": "npm run generate:font-regular && npm run generate:font-filled && npm run generate:font-resizable",
+    "generate:rtl": "node ../../importer/rtlMetadata.js --source=../../assets --dest=./intermediate",
     "rollup": "node ./generateRollup.js",
     "optimize": "svgo --config svgo.config.js --folder=./intermediate --precision=2",
     "unfill": "find ./intermediate -type f -name \"*.svg\" -exec sed -i.bak 's/fill=\"none\"//g' {} \\; && find ./intermediate -type f -name \"*.bak\" -delete",
-    "build": "npm run copy && npm run generate:font && npm run optimize && npm run unfill && npm run convert:svg && npm run convert:fonts && npm run cleanSvg && npm run build:esm && npm run build:cjs && npm run copy:font-files",
+    "build": "npm run copy && npm run generate:font && npm run generate:rtl && npm run optimize && npm run unfill && npm run convert:svg && npm run convert:fonts && npm run cleanSvg && npm run build:esm && npm run build:cjs && npm run copy:font-files",
+    "buildSvg": "npm run copy && npm run generate:rtl && npm run optimize && npm run unfill && npm run convert:svg && npm run cleanSvg && npm run build:esm && npm run build:cjs",
```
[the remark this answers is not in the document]
tomi-msft: It's for testing locally, but I figured I would leave it there so I wouldn't have to keep adding it anytime I need to test just the svgs

importer/rtlMetadata.js
```diff
@@ -0,0 +1,109 @@
...
+          } else if (file === 'metadata.json') {
+            // If it's a metadata file, read and parse its content
+            fs.readFile(srcFile, 'utf8', (err, data) => {
+                if (err) {
+                    console.error('Error reading metadata file:', err);
+                    return;
+                }
+                try {
+                    // Parse the json content
+                    let metadata = JSON.parse(data);
```
spencer-nelson, line 61: For the metadata parsing, following this specific pattern for the name in the resultant json won't work for platforms that don't follow this pattern. Can the name processing be part of the platform processing, rather than the general parts here so this can be more broadly useful?

spencer-nelson (approved):

PR #639, opened 2023-09-11
Update iOS importer to use new metadata

nickromano:
Support the new metadata.json for locales and LTR/RTL support on iOS


williamch-ms:
Noticed the pr.yml file has a step that calls for running deploy:ios. Just want to verify with you that we won't be running publishing on the pull request pipeline step..

https://github.com/microsoft/fluentui-system-icons/blob/main/.github/workflows/pr.yml#L62

nickromano:
@williamch-ms Yes that just tests that the scripts to create the iOS library work. It won't actually publish the library until the build is tagged.

spencer-nelson (approved):

PR #679, opened 2024-01-09
Fix bundleIcon primaryFill behavior

ReillyBova:
The bundleIcon util is not successfully carrying the primaryFill prop down to its regular or filled child. The `fill` prop it is passing the primaryFill value to does not actually apply a color change as of the latest package version, tested on Chrome/macOS with the latest stable version of Fluent v9.

See this issue:

https://github.com/microsoft/fluentui/issues/30272


tomi-msft (approved):
approved. thanks for making the fix

PR #721, opened 2024-05-28
Script updates for tooling

willchavez:
- diff.sh, now also downloads the previous versions of the updates file, and uses imagemagick to compare them.
- Generate HTML Report, python script that creates a replica of the visual-diff tool we use for icon updates

icon_updates


scripts/diff.sh
```diff
@@ -0,0 +1,142 @@
...
+# Ensure the script is run from the correct directory
+# Do not delete this check, this scripts needs to be ran at the root of the repo.
+if [[ $(basename "$PWD") == "scripts" ]]; then
+    echo "Please run this script from the root directory of your repository."
+    exit 1
+fi
+
+# Prompt the user for the commit IDs
+read -p "Enter the commit ID from (inclusive): " commit_id_from
+read -p "Enter the commit ID to: " commit_id_to
```
spencer-nelson, line 27: Is it possible to just run this with a from and no "to" so it runs to the current commit?

scripts/diff.sh
```diff
@@ -0,0 +1,142 @@
+#!/bin/bash
+
+# This script extracts files changed between two specified Git commit IDs and performs various file operations.
```
spencer-nelson, line 3: For this to work you have to be on a branch that has the specific commit ids as part of the branch history. Sometimes it just requires a git pull, sometimes you might have to check out another branch, but that's worth noting as a possible error.

scripts/generate_html_report.py
```diff
@@ -0,0 +1,239 @@
...
+
+Note:
+- The HTML report includes buttons to toggle the visibility of new icons and to zoom in/out on the icons.
+"""
+
+import os
+import re
+
+# Prompt the user for the path to the SVGs folder
+svg_folder = input("Please enter the path to the SVGs folder: ")
```
spencer-nelson, line 39: can there be a default value to this so one can just hit enter and not have to add a path?

spencer-nelson (changes requested):
The default values are a nicety, but I feel it's critical to include instructions about being on a valid branch. I've been caught by that too many times.

willchavez:
@spencer-nelson just made some updates, excellent suggestions. 💯

spencer-nelson (approved):

PR #729, opened 2024-06-18
Add CODEOWNERS file

tomi-msft:
Add a CODEOWNERS file


spencer-nelson:
This is a great idea. Can you please further flesh this out so the other platform contributors as well as the design maintainers are also included to the relevant directories? At a minimum, Nick Romano should be included for iOS, PraveenKumar Yeruva for Android, Will and myself for assets. For the importer directory, might be worth adding everyone so they at least get the extra notifications.

nickromano (approved):

PR #807, opened 2025-03-08
Reduce size of package brought in by react-native-icons to RN bundles

rurikoaraki:
We would like to support using this package in React Native scenarios in order to pull icons from a centralized place.
Unfortunately, due to how the exports of the package are currently set up, the package ends up pulling in information about all the chunks exported by the package and adding a lot of bloat. Specifically, `export *` causes problems for esbuild's tree shaking, so instead we should be exporting all the icons by name instead.

Additionally, some of the logic and typings in the package are not necessary for React Native scenarios or do not align with React Native usage, so I am fixing those up too. This also allows removing the griffel styling package and using PlatformColors, needed for HC scenarios.

Reduction in package size that gets pulled in from adding the @fluentui/react-native-icons package from what I've seen in the release bundle is about 30KB.

### DIFF
--- a/packages/react-native-icons/src/utils/FluentIconsProps.types.ts
+++ b/packages/react-native-icons/src/utils/FluentIconsProps.types.ts
@@ -1,8 +1,8 @@
-import * as React from "react";
+import type { ColorValue, TextProps } from "react-native";
+import type { SvgProps } from "react-native-svg";
 
-export type FluentIconsProps<TBaseAttributes extends (React.SVGAttributes<SVGElement> | React.HTMLAttributes<HTMLElement>) = React.SVGAttributes<SVGElement>> = (TBaseAttributes) & {
-	primaryFill?: string
-	className?: string
+export type FluentIconsProps<TBaseAttributes extends (SvgProps | TextProps) = SvgProps> = (TBaseAttributes) & {
+	primaryFill?: ColorValue
 	filled?: boolean
 	title?: string
 }
--- a/packages/react-native-icons/src/utils/fonts/createFluentFontIcon.tsx
+++ b/packages/react-native-icons/src/utils/fonts/createFluentFontIcon.tsx
@@ -1,23 +1,28 @@
 import * as React from 'react';
-import { FluentIconsProps } from '../FluentIconsProps.types';
-import { makeStyles, makeStaticStyles, mergeClasses } from "@griffel/react";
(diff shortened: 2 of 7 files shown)


packages/react-native-icons/src/utils/fonts/createFluentFontIcon.tsx
```diff
@@ -26,89 +31,83 @@ export const enum FontFile {
...
-    root: {
-        display: 'inline-block',
-        fontStyle: 'normal',
-        lineHeight: '1em',
-
-        "@media (forced-colors: active)": {
-            forcedColorAdjust: 'auto',
+// const FONT_FAMILY_MAP = {
+//     [FontFile.Filled]: 'FluentSystemIconsFilled',
+//     [FontFile.Regular]: 'FluentSystemIconsRegular',
```
Saadnajmi, line 36: Remove all the commented code?

rurikoaraki: I can, I was going to leave it for later for when we can do font icon for RN, but I guess I can just look at what the react-icons package has

Saadnajmi (approved):

packages/react-native-icons/src/utils/FluentIconsProps.types.ts
```diff
@@ -1,8 +1,8 @@
-import * as React from "react";
+import type { ColorValue, TextProps } from "react-native";
+import type { SvgProps } from "react-native-svg";
 
-export type FluentIconsProps<TBaseAttributes extends (React.SVGAttributes<SVGElement> | React.HTMLAttributes<HTMLElement>) = React.SVGAttributes<SVGElement>> = (TBaseAttributes) & {
-	primaryFill?: string
-	className?: string
```
spencer-nelson, line 5: Is there any way removing the className might cause problems for folks, or what is the likelihood it was being used by anyone? Alternately, is there a significant cost to leaving it in so as to prevent possible negative impact to users?

spencer-nelson (approved):

packages/react-native-icons/src/utils/FluentIconsProps.types.ts
rurikoaraki: Also we don't use classNames in RN, so removing them shouldn't be an issue. Styles get added via passing a property bag into a style prop for RN, we don't have CSS classNames we can use, unfortunately

PR #842, opened 2025-06-16
Add Compose Multiplatform library

terrakok:
with a demo app

https://github.com/user-attachments/assets/10ba926f-1081-47a7-acf2-ea76010b996b


terrakok:
@etudie could you take a look?

PR #855, opened 2025-07-02
Remove forcedColorAdjust: 'auto' from font icons

### LINKED ISSUE #852: V9 Primary Button's and Split Button's font icon is not visible in high contrast (aquatic and desert) modes
Screenshot showing issue:
Image
Image

If forceColorAdjust is set to be 'none', it will help fix the issue.

**Related PR checked in earlier that impacts the scenario:**
There was a PR checked in changed the forceColorAdjust from 'none' to 'auto', [fix: Add styling to font icons to fix HC color issues (#781) · microsoft/fluentui-system-icons@e01fd76](https://github.com/microsoft/fluentui-system-icons/commit/e01fd7683c737ad844faddd572c427669ddf9559)

emmayjiang:
Changes the `forced-colors-adjust` value from `auto` / `none` to the default value, `inherit`, in hopes of addressing the issues with backplates. Also leaves a explanation for why we took this approach, in case it needs to be re-evaluated in the future.

Fixes #852

### DIFF
--- a/packages/react-icons/src/utils/fonts/createFluentFontIcon.tsx
+++ b/packages/react-icons/src/utils/fonts/createFluentFontIcon.tsx
@@ -72,9 +72,15 @@ const useRootStyles = makeStyles({
         lineHeight: '1em',
         color: 'currentColor',
 
-        "@media (forced-colors: active)": {
-            forcedColorAdjust: 'auto',
-        }
+        // Sometimes you need the font icon to have forced-color-adjust: none to not have a
+        // backplate, which should generally follow the needs of the parent control.
+
+        // But sometimes you need forced-color-adjust: auto, especially when teams have added
+        // a non-high-contrast color style to the icon's styles that should be automatically
+        // overridden in WHCM.
+
+        // Rather than setting forced-color-adjust to 'none' or 'auto', we leave this value
+        // default, inherited from the parent control.
     },
     [FontFile.Filled]: {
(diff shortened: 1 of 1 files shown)


packages/react-icons/src/utils/createFluentIcon.ts
```diff
@@ -12,11 +13,22 @@ export type CreateFluentIconOptions = {
     color?: boolean;
 }
 
+const useRootStyles = makeStyles({
```
smhigley, line 16: Maybe one comment here too, about how this is because browsers automatically have SVGs set to `forced-color-adjust: preserve-parent-color`, which will not override internally-defined colors in our SVG icons in high contrast mode.

smhigley (approved):

vinguyen12:
> Hi Emma. Thank you for raising this fix!
> 
> We realized that force-color-adjust: auto are initialized in 2 places, (1) one is in packages/react-icons/src/utils/fonts/createFluentFontIcon.tsx (the one mentioned in the fix) and (2) the other one is in [fluentui/packages/react-components/react-button/library/src/components/ToggleButton/useToggleButtonStyles.styles.ts](https://github.com/microsoft/fluentui/blame/master/packages/react-components/react-button/library/src/components/ToggleButton/useToggleButtonStyles.styles.ts). Therefore, the issue is still reproducing after the change.

It is fixed with latest commit

PR #858, opened 2025-07-15
feat(react-icons): enable griffel unprocessed (styles.raw.js) output for production assets

Hotell:
Adds `.styles.raw.js` output to production assets. This enables opt in transpilation which avoids known issue with griffel style clashes.

- moves all transpilation related logic to `build.js`
- adds additional integration tests to verify `styles.raw.js` is produced and non `styles.js` is processed by griffel
- decouples all griffel related code to `.styles.ts` files 
  - this will be enforced via lint as a follow-up 

---

## Related issues: 

- https://github.com/microsoft/fluentui/pull/34780

### DIFF
--- a/packages/react-icons/build.js
+++ b/packages/react-icons/build.js
@@ -11,18 +11,24 @@ const { transformSync } = require('@babel/core');
 
 main();
 
+/**
+ * Builds source TypeScript and copys assets to the output directories.
+ *
+ * It transpiles TypeScript files to both ESNext and CommonJS formats,
+ * applies Babel transformations, and copies font assets.
+ * It also creates raw style copies for .styles.js files.
+ */
 function main(){
-  const projectRoot = join(__dirname);
+  const projectRoot = __dirname;
 
-  transpileTsc(projectRoot, { moduleFormat: 'esnext', outDir: 'lib' });
-  transpileTsc(projectRoot, { moduleFormat: 'commonjs', outDir: 'lib-cjs' });
+  transpileTsc({ moduleFormat: 'esnext', outDir: 'lib' },projectRoot);
+  transpileTsc({ moduleFormat: 'commonjs', outDir: 'lib-cjs' },projectRoot);
 
   applyBabelTransform('lib', projectRoot);
   applyBabelTransform('lib-cjs', projectRoot);
 
-  const assets = glob.sync('src/utils/fonts/*.{ttf,woff,woff2,json}', { cwd: projectRoot });
-  copyAssets(assets, './lib/utils/fonts', projectRoot);
-  copyAssets(assets, './lib-cjs/utils/fonts', projectRoot);
+  copyAssets('src/utils/fonts/*.{ttf,woff,woff2,json}', './lib/utils/fonts', projectRoot);
+  copyAssets('src/utils/fonts/*.{ttf,woff,woff2,json}', './lib-cjs/utils/fonts', projectRoot);
 
 }
 
@@ -49,10 +55,10 @@ function createRawStylesCopy(styleFile) {
 
 /**
  *
- * @param {string} baseDir
  * @param {{moduleFormat:'esnext'|'commonjs'; outDir:string}} options
+ * @param {string} baseDir
  */
-function transpileTsc(baseDir,options) {
+function transpileTsc(options, baseDir) {
  const {moduleFormat,outDir} = options;
   console.log(`Transpiling module format [${moduleFormat}] to -> ${outDir}/`);
   const cmd = `npx tsc -p . --module ${moduleFormat} --outDir ${outDir}`;
@@ -113,15 +119,16 @@ function applyBabelTransform(root, baseDir) {
 
 /**
  *
- * @param {string[]} src
+ * @param {string} src - globs of files to copy
  * @param {string} dest
  * @param {string} baseDir
  */
 function copyAssets(src, dest, baseDir){
 
+  const assets = glob.sync(src, { cwd: baseDir });
+  console.log(`Copying ${src} assets to -> ${dest}:`);
 
-  console.log(`Copying assets to ${dest}:`);
-  src.forEach(file => {
+  assets.forEach(file => {
     const sourcePath = join(baseDir, file);
     const targetPath = join(baseDir,dest, basename(file));
 


packages/react-icons/src/utils/fonts/createFluentFontIcon.shared.ts
```diff
@@ -0,0 +1,9 @@
+export const enum FontFile {
+    Filled = 0,
+    Regular = 1,
+    Resizable = 2,
+	Light = 3
+}
+
```
dmytrokirpa, line 7: Is prettier configured for this project?

packages/react-icons/src/utils/fonts/createFluentFontIcon.styles.ts
```diff
@@ -0,0 +1,83 @@
+
+import { makeStyles, makeStaticStyles, mergeClasses } from "@griffel/react";
```
dmytrokirpa: nit: `mergeClasses` isn't used in this file

dmytrokirpa (approved):
Approving to unblock the release, but we need to follow-up

packages/react-icons/src/utils/fonts/createFluentFontIcon.shared.ts
Hotell: no

Hotell:
> Approving to unblock the release, but we need to follow-up on tools/infra - eg formatter, refactor build scripts, etc

yes https://github.com/microsoft/fluentui-system-icons/pull/859

and additional items will need follow ( but ATM low prio )

packages/react-icons/src/utils/fonts/createFluentFontIcon.shared.ts
Hotell: i avoided these things on purpose as the diff created would create lot of churn which is not wanted for this feature

packages/react-icons/tsconfig.json
```diff
@@ -1,9 +1,11 @@
 {
   "compilerOptions": {
+    "pretty": true,
     "outDir": "lib",
     "target": "ES2019",
     "module": "esnext",
     "lib": ["ES2019", "dom"],
+    "types": [],
```
Hotell: following best practices, we don't wanna all repo `@types/*` be included for `tsc` invocation ( this will also fail the CI because vitest adds additional baggage )

packages/react-icons/build.js
```diff
@@ -0,0 +1,136 @@
...
+
+main();
+
+function main(){
+  const projectRoot = join(__dirname);
+
+  transpileTsc(projectRoot, { moduleFormat: 'esnext', outDir: 'lib' });
+  transpileTsc(projectRoot, { moduleFormat: 'commonjs', outDir: 'lib-cjs' });
+
+  applyBabelTransform('lib', projectRoot);
```
Hotell, line 27: need to keep re applying babel on all tsc outputs to keep consistent output - we can change this as a follow up - for this task it's out of scope for now

packages/react-icons/package.json
```diff
@@ -14,7 +14,6 @@
   "scripts": {
     "clean": "git clean -fXd lib/ lib-cjs/ intermediate/ src/",
     "copy": "node ../../importer/generate.js --source=../../assets --dest=./intermediate --extension=svg --target=react",
-    "copy:font-files": "cpy './src/utils/fonts/*.{ttf,woff,woff2,json}' ./lib/utils/fonts/. && cpy './src/utils/fonts/*.{ttf,woff,woff2,json}' ./lib-cjs/utils/fonts/.",
```
Hotell, line 17: covered by build.js

packages/react-icons/package.json
```diff
@@ -25,23 +24,18 @@
     "generate:rtl": "node ../../importer/rtlMetadata.js --source=../../assets --dest=./intermediate/rtl.json",
     "optimize": "svgo --config svgo.config.js --folder=./intermediate --precision=2 --quiet",
     "unfill": "find ./intermediate -type f -name \"*.svg\" ! -name \"*_color*\" -exec sed -i.bak 's/fill=\"none\"//g' {} \\; && find ./intermediate -type f -name \"*.bak\" -delete",
-    "build:cjs": "tsc -p . --module commonjs --outDir lib-cjs && babel lib-cjs --out-dir lib-cjs",
-    "build:esm": "tsc -p . && babel lib --out-dir lib",
```
Hotell, line 29: covered by build.js

packages/react-icons/package.json
```diff
@@ -25,23 +24,18 @@
...
-    "build:js": "npm run build:cjs && npm run build:esm",
+    "build:js": "node build.js",
     "prebuild": "npm run clean",
-    "build": "npm run build:fonts-and-svg && npm run build:generate-js-chunks && npm run build:js && npm run copy:font-files",
+    "build": "npm run build:fonts-and-svg && npm run build:generate-js-chunks && npm run build:js",
     "build:verify": "vitest run build-verify.test.js",
     "test": "vitest --exclude build-verify.test.js"
   },
   "devDependencies": {
-    "@babel/cli": "^7.16.0",
```
Hotell, line 39: unused deps

packages/react-icons/.babelrc
```diff
@@ -1,4 +1,3 @@
 {
-  "presets": ["@griffel"],
```
Hotell, line 2: griffel is applied only on `.styles.ts` programatically within build.js

packages/react-icons/build-verify.test.js
```diff
@@ -133,10 +133,88 @@ describe('Build Verification', () => {
     });
   });
 
- // TODO: enable once griffel RAW modules are available
-  describe.skip('Styles Files', () => {
-    it('should have required styles files in utils', async () => {
-      const utilsPath = path.join(__dirname, 'lib', 'utils');
+  describe('Styles Files', () => {
+   it(`should produce griffel processed .styles.js and unprocessed .styles.raw.js [lib]`, () => {
```
Hotell, line 137: integration snapshots to verify griffel works as expected

packages/react-icons/build-verify.test.js
```diff
@@ -133,10 +133,88 @@ describe('Build Verification', () => {
     });
   });
 
- // TODO: enable once griffel RAW modules are available
```
Hotell, line 136: enabling as raw is in

packages/react-icons/build.js
```diff
@@ -0,0 +1,143 @@
...
+    const filePath = join(jsRoot, filename);
+
+    if (isStylesFile) {
+      createRawStylesCopy(filePath)
+    };
+
+    const codeBuffer = readFileSync(filePath);
+    const sourceCode = codeBuffer.toString().replace(EOL_REGEX, '\n');
+
+    const result = transformSync(sourceCode, {
```
dmytrokirpa, line 96: do we need to post-process all files with babel or only `styles.(raw.)?js`?

Hotell: to get same transpilation output yes, initially in this PR i applied babel only on `.styles` but after addition of thorough integration tests they started to fail because the output from `generate chunks --> tsc --> babel(on .styles)` is meesing with the `PURE` annotations and completely removes them in `cjs`, thus to keep it as is I reverted that approach and now every js file is run throught babel but only `.styles` are using griffel transform

PR #859, opened 2025-07-15
feat(react-icons,react-native-icons): improve build performance

Hotell:
needs: https://github.com/microsoft/fluentui-system-icons/pull/858

| task | Before | After |
|--------|--------|--------|
| `react-icons build` | 6m 43s | 1m 54s / 𝚫 72% FASTER |
| `react-native-icons build` | 5m 51s | 2m 12s / 𝚫 63% FASTER |

## general

- leverages task concurrency where possible
- makes build platform agnostic
- encapsulate resource heavy shell `unfill` task to nodejs script
- updates GHA(CI) to use secure version of actions + migrates to Node v20

## react-icons

- source files are transpiled via `tsc` for both esm and cjs
- generated content: 
  - whole chunks output is completely generated, including `.d.ts` files, to overcome huge perf implications
  - type definitions are encapsulated into single definition to make `tsc` processing faster 
  - tsc@5.6 is executed via `--no-check` mode to transpile generated esm (.js) files to cjs

## react-native-icons

- removes babel step completely as `tsc` handles jsx transpilation as expected


.github/CODEOWNERS
```diff
@@ -1,6 +1,6 @@
 ## packages
 packages/icon-app @tomi-msft
-packages/react-icons @tomi-msft @behowell
+packages/react-icons @microsoft/cxe-prg @microsoft/cxe-red
```
spencer-nelson, line 3: Is there a reason we shouldn't reflect this change for the other packages that were being maintained by @tomi-msft?

packages/react-icons/unfill-svg.js
```diff
@@ -0,0 +1,44 @@
...
+      if (entry.isFile() &&
+          entry.name.endsWith('.svg') &&
+          !entry.name.includes('_color')) {
+
+        const content = await fs.readFile(fullPath, 'utf8');
+        const updated = content.replace(/fill="none"/g, '');
+
+        if (content !== updated) {
+          await fs.writeFile(fullPath, updated);
+          console.log(`Processed: ${fullPath}`);
```
spencer-nelson, line 32: Are you sure you want a console log for every file here? This pattern is in various places of the pr and will greatly bloat the output log

.github/CODEOWNERS
Hotell: probably in time, not for now as that's out of scope

PR #861, opened 2025-07-21
chore: add build output integration tests to react-icons, migrate ci to node v20 and secure github actions

Hotell:
This pull request introduces updates to workflows, TypeScript configurations, and package scripts to modernize dependencies, improve build processes, and enhance testing capabilities. Key changes include upgrading Node.js and GitHub Actions versions, adding TypeScript type definitions, and integrating `vitest` for testing.

### Workflow Updates:
* Updated `.github/workflows/pr.yml` and `.github/workflows/publish.yml` to use `actions/checkout@v4` and `actions/setup-node@v4` with Node.js 20.x. Replaced `npm install` with `npm ci` for faster and more reliable dependency installation. [[1]](diffhunk://#diff-15c806aa509538190832852f439e9921a23bec2da81f95ed0e4bf13c14e5b160L11-R11) [[2]](diffhunk://#diff-15c806aa509538190832852f439e9921a23bec2da81f95ed0e4bf13c14e5b160L52-R58) [[3]](diffhunk://#diff-551d1fcf87f78cc3bc18a7b332a4dc5d8773a512062df881c5aba28a6f5c48d7L32-R32) [[4]](diffhunk://#diff-551d1fcf87f78cc3bc18a7b332a4dc5d8773a512062df881c5aba28a6f5c48d7L63-R67)
* Added new build and test steps in `pr.yml` for `@fluentui/react-icons` and `@fluentui/react-icons-font-subsetting-webpack-plugin`.

### TypeScript and Build Configuration Updates:
* Modified `tsconfig.json` files across packages (`react-icons`, `react-native-icons`, `react-icons-font-subsetting-webpack-plugin`) to include `types` field and exclude `node_modules` from compilation. [[1]](diffhunk://#diff-004f4c36d4d33930750c6e2d2e04098da6bfb4201a27ad88f3aadbcf01d19a6aR13) [[2]](diffhunk://#diff-5e03ef87e1a33a7843a4fd683303306984ae628b161513d2b1980e91447a4fc8L18-R20) [[3]](diffhunk://#diff-001e1f6f8212ef8292b29e53f0b1c02972b3f7b61ab8d5404f36a45735d7f3aeL29-R30)
[...]

PR #862, opened 2025-07-22
fix(react-icons): make font svg generation deterministic

Hotell:
This pull request improves the `processFolder` function in `packages/react-icons/convert-font.js` to ensure deterministic chunk generation and consistent processing order for icon files.

### Improvements to file processing:

* Updated the file selection logic to explicitly specify the order of icon files (`Light`, `Regular`, `Filled`) for deterministic chunk generation. This replaces the glob pattern with a predefined file list and filters for existing files.
* Changed the processing of files from parallel (`Promise.all`) to sequential iteration (`for...of`) to maintain a consistent order when generating icon entries and codepoint maps.

## Related issues

- follows https://github.com/microsoft/fluentui-system-icons/pull/861

### DIFF
--- a/packages/react-icons/convert-font.js
+++ b/packages/react-icons/convert-font.js
@@ -87,27 +87,33 @@ async function processFiles(src, dest) {
 
 /**
  * Process a folder of svg files and convert them to React components, following naming patterns for the FluentUI System Icons
- * @param {string} srcPath 
+ * @param {string} srcPath
  * @param {string} codepointMapDestFolder
- * @param {boolean} resizable 
+ * @param {boolean} resizable
  * @returns { Promise<string[]> } - chunked icon files to insert
  */
 async function processFolder(srcPath, codepointMapDestFolder, resizable) {
-  var files = await glob(resizable ? 'FluentSystemIcons-Resizable.json' : 'FluentSystemIcons-{Filled,Regular,Light}.json', { cwd: srcPath, absolute: true });
+
+  // Explicitly specify the order to make chunk generation deterministic: Light first, then Regular, then Filled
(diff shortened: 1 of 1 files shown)


packages/react-icons/convert-font.js
```diff
@@ -87,27 +87,33 @@ async function processFiles(src, dest) {
...
 async function processFolder(srcPath, codepointMapDestFolder, resizable) {
-  var files = await glob(resizable ? 'FluentSystemIcons-Resizable.json' : 'FluentSystemIcons-{Filled,Regular,Light}.json', { cwd: srcPath, absolute: true });
+
+  // Explicitly specify the order to make chunk generation deterministic: Light first, then Regular, then Filled
+  const fileNames = resizable ? ['FluentSystemIcons-Resizable.json'] : ['FluentSystemIcons-Light.json', 'FluentSystemIcons-Filled.json', 'FluentSystemIcons-Regular.json'];
+  const files = fileNames.map(name => path.resolve(srcPath, name)).filter(filePath => fsS.existsSync(filePath));
 
   /** @type string[] */
   const iconExports = [];
-  await Promise.all(files.map(async (srcFile, index) => {
```
Hotell, line 100: this was causing different chunk output on every build

dmytrokirpa: even with pre-defined files order?

dmytrokirpa: also how does it affect the build time?

Hotell: not 100% sure about the previous glob return order but that might provided different order + the async iterator could feed the `iconExports` array differently.

checked shipped assets and it seems `esm` have been not affected but `cjs` has been ( which is clearly the case https://github.com/microsoft/fluentui-system-icons/actions/runs/16443486508/job/46469508540?pr=858).

PR #868, opened 2025-07-23
chore(react-icons): improve build performance

Hotell:
- moves unfiling functionality to nodejs cli
  - including `--dry-run` mode
  - 100x faster than previous `sed` bash solution
- runs font generation in parallel

## Performance

| task | Before | After |
|--------|--------|--------|
| `react-icons build` | 4m 1s | 2m 27s / 𝚫 39% FASTER |
| `npm run unfill` | 2m | 2s / 𝚫 99% FASTER |

---
Replaces https://github.com/microsoft/fluentui-system-icons/pull/859

### DIFF
--- a/packages/react-icons/package.json
+++ b/packages/react-icons/package.json
@@ -13,18 +13,18 @@
   },
   "scripts": {
     "clean": "git clean -fXd lib/ lib-cjs/ intermediate/ src/",
-    "copy": "node ../../importer/generate.js --source=../../assets --dest=./intermediate --extension=svg --target=react",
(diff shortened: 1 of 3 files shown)


packages/react-icons/package.json
```diff
@@ -13,18 +13,18 @@
   },
   "scripts": {
     "clean": "git clean -fXd lib/ lib-cjs/ intermediate/ src/",
-    "copy": "node ../../importer/generate.js --source=../../assets --dest=./intermediate --extension=svg --target=react",
+    "generate:assets-to-svg": "node ../../importer/generate.js --source=../../assets --dest=./intermediate --extension=svg --target=react",
```
Hotell, line 16: renamed to more appropriate name to express the intent

dmytrokirpa (approved):

PR #873, opened 2025-07-25
Icon app/make it work

Hotell:


packages/icon-app/package.json
```diff
@@ -5,18 +5,21 @@
...
   },
   "author": "",
   "license": "ISC",
   "devDependencies": {
     "@types/react": "^17.0.2",
     "@types/react-dom": "^17.0.0",
-    "html-webpack-plugin": "^5.5.0",
+    "html-webpack-plugin": "^5.6.3",
     "ts-loader": "^9.2.9",
-    "typescript": "^4.6.3",
```
Hotell, line 17: bumpt to ts 4.7. which has moduleResolution: bundle -> resolves types against export maps

Issue #788, opened 2024-12-17
Make icon metadata available via a package


micahgodbolt:
I'm looking to use the icons meta data while rendering our docs, specifically to allow users to search by metaphor to find the icon. It doesn't make sense for each package to separately publish this, so is this something that could be created as its own package?

willchavez:
Hey @micahgodbolt can you expand on this? Would a separate package with all the metadata compressed into a single file fit your need? Or separate json files fine?

micahgodbolt:
yeah, a single file would be fine. I'd love to be able to take an icon like AccessibilityErrorFilled, and look up all of the related terms. This would greatly improve search, as we can only currently search by exact name.

https://react.fluentui.dev/?path=/docs/icons-catalog--docs

![Image](https://github.com/user-attachments/assets/0af1284d-b2f5-45f9-85c2-4b4124859816)

MannGo1:
Hi Fluent Team. we also have a requirement where we give user ability to assign an icon corresponding to a property displayed on a card. For this user needs to search the icons. We were thinking to use the metaphors associated each icon from the icon's metadata.json file. But since these metaphors are not accessible programatically we would have to download the metadata.json and use that for each icon. So we would also like to have these feature, so that we can access these metaphors using a pacakge.

PR #876, opened 2025-07-28
chore(react-icons): add metadata.json generation and validation

Hotell:
## Feat: Adds generation of `metadata.json`  to reflect icons being shipped to npm.

ATM this is not part of publish pipeline, which is desirable in order to prevent any type of Breaking Changes for our consumers. In order to incorporate this within publish process the infra would need to change
- GHA cannot push to protected branches --> will need to migrate to same release process as in Fluent(contrib)-> manually triggered from brach/Release-PR

**metadata format:**

> see current output https://raw.githubusercontent.com/microsoft/fluentui-system-icons/604b8d627951fd40bab650e0983f6b1a34a4921a/packages/react-icons/metadata.json

```jsonc
{
    "IconName": {
      "svg": boolean,    // true if icon exists in SVG format
      "font": boolean,   // true if icon exists in font format
      "type": "resizable" | "sized"  // icon type classification
    }
  }
```

**metadata validation:**

- metadata.json is stored in repo
- part of `npm run build:verify` executed in `vitest` context / https://github.com/microsoft/fluentui-system-icons/pull/876/files#diff-a13e3db696ed7beec1767d89893d47098e50e046f3e55cf129c9c800ee0cada0R1565

## Related issues

- https://github.com/microsoft/fluentui-system-icons/issues/788

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,6 +36,7 @@ packages/react-icons/src/utils/fonts/*.woff2
 packages/react-icons/src/utils/fonts/*.css
 packages/react-icons/src/utils/fonts/*.html
 packages/react-icons/src/utils/fonts/*.json
+packages/react-icons/src/utils/fonts/**/*.svg
 
 packages/react-native-icons/intermediate
 packages/react-native-icons/react-icons-test/


packages/react-icons/convert-font.js
```diff
@@ -154,12 +177,14 @@ async function generateCodepointMapForWebpackPlugin(destPath, iconEntries, resiz
...
 function generateReactIconEntries(iconEntries, resizable) {
   /** @type {string[]} */
   const iconExports = [];
-  const metadata = JSON.parse(fsS.readFileSync(RTL_FILE, 'utf-8'));
+  /** @type {string[]} */
+  const iconNames = [];
+  const metadata = JSON.parse(fsS.readFileSync(/** @type {string} */ (RTL_FILE), 'utf-8'));
   for (const [iconName, codepoint] of Object.entries(iconEntries)) {
     let destFilename = getReactIconNameFromGlyphName(iconName, resizable);
     var flipInRtl = metadata[destFilename] === 'mirror';
```
dmytrokirpa, line 191: nit, and might be out of scope for the PR, but:

```suggestion
    const flipInRtl = metadata[destFilename] === 'mirror';
```

packages/react-icons/merge-metadata.js
```diff
@@ -0,0 +1,111 @@
...
+      console.warn(`Warning: Could not read font metadata from ${fontMetadataPath}:`, error.message);
+    }
+  }
+
+  // Merge metadata: combine SVG and font data
+  const mergedMetadata = {};
+
+  // Add all SVG icons
+  for (const [iconName, iconData] of Object.entries(svgMetadata)) {
+    mergedMetadata[iconName] = { ...iconData };
```
dmytrokirpa: You can get rid if the loop and initialize the `mergedMetadata` with a deep copy of `svgMetadata`

```ts
  const mergedMetadata = structuredClone(svgMetadata);
```

packages/react-icons/convert-font.js
Hotell: yeah I plan to clean things up after "in review" prs land in a batch
- adding prettier
- modernizing the scripts etc

packages/react-icons/merge-metadata.js
Hotell: good catch, wondering which approach is faster, the new native APIs tend to be slower in general

dmytrokirpa (approved):

PR #877, opened 2025-08-04
chore(react-icons): add bundle-size metrics,gates

Hotell:
Adds bundle size measurement and reporting for monorepo via `monosize` tool.

1. new PR workflow step 
  - runs `monosize measure` for existing fixtures 
  - runs `monosize compare-reports` comparing repo stored storage vs current PR measurement

image

2. new baseline workflow

- runs on merge to main if react-icons or assets have been modified in order to update baseline metrics

3. implements custom GitHub `createLocalStorage` monosize adapter which operates on physical repo files  

- compares PR against baseline stored in default branch  
- stores baseline via default branch main pipeline to repo 

4. nx

- setups cache for `build` target on `react-icons`

### DIFF
--- a/.github/workflows/bundle-size.baseline.yml
+++ b/.github/workflows/bundle-size.baseline.yml
@@ -25,13 +25,13 @@ jobs:
           app-id: ${{ secrets.GH_FLUENT_CI_APP_ID }}
           private-key: ${{ secrets.GH_FLUENT_CI_PRIVATE_KEY }}
 
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v6
         with:
           token: ${{ steps.generate-token.outputs.token }}
 
       - uses: actions/setup-node@v4
         with:
-          node-version: 20.x
+          node-version: 22.x
 
       - run: npm ci
 


dmytrokirpa (dismissed):

packages/react-icons/monosize.config.mjs
```diff
@@ -0,0 +1,66 @@
...
+  threshold: '10kB',
+};
+
+export default config;
+
+/**
+ * Creates a local storage adapter for Monosize.
+ * This adapter reads and writes bundle size reports to a local JSON file.
+ * It is used to store the bundle size reports locally for the package.
+ * TODO: needs to be reworked to leverage GHA artifacts because pushing from CI to protected branch is not allowed.
```
Hotell: this should no longer be the case, will remove once validated after merge

packages/react-icons/BUNDLE-SIZE-ANALYSIS.md
```diff
@@ -0,0 +1,7 @@
+**M365 Copilot (data with react-icons alpha)**
```
dmytrokirpa: This should probably be removed

dmytrokirpa (approved):

PR #878, opened 2025-08-05
test(react-icons): remove transpiler commonjs void export.* churn from snapshots and sync after latest release

Hotell:
This PR implements a truncating mechanism for these CJS export void 0 statements emmitted by transpiler to mitigate churn for diff and making them more manageable

## Why is this needed?

Latest release added/removed some icons leading into following chunk diff in CJS output:

**Added exports:**

exports.CalendarCheckmarkSparkleRegular
exports.CalendarCheckmarkSparkleFilled

**Missing from expected:**

> These exists but in different chunk

exports.Cellular3GRegular
exports.Cellular3GFilled


packages/react-icons/build-verify.test.js
```diff
@@ -9,15 +9,35 @@ import { execSync } from 'node:child_process';
...
-  const trimmedLines = lines.slice(0, 30);
 
-  if (lines.length > 30) {
+  // Count and filter out verbose CJS export declarations to reduce churn
+  const voidExportLines = [];
+  const filteredLines = lines.filter(line => {
+    // Check for lines that are just "exports.SomeName = void 0;" declarations
+    const isVoidExport = /^\s*exports\.[a-zA-Z0-9_]+ = .*void 0;\s*$/.test(line);
+    if (isVoidExport) {
+      voidExportLines.push(line);
```
dmytrokirpa: Not a big deal, but I don't really like side-effects in predicates, not sure if it would be better to use for-of with two pushes instead, wdy?

dmytrokirpa (approved):

packages/react-icons/build-verify.test.js
Hotell: yeah to make this super clean a `reduce` would be appropriate, lemme fix

Hotell: https://github.com/microsoft/fluentui-system-icons/pull/878/commits/4e5418c033d5426406176aa4bb188205cf1f679e

dmytrokirpa (approved):

PR #907, opened 2025-10-16
chore: define nx target dependencies, resolve fantasticon race conditions and make raw tests work

Hotell:
- define nx target dependencies
  - `nx affected -t build` will fail on CI because there is no relationship specified ( icon-app needs react-icons:build first )
- resolve fantasticon  race condition while patching
  - issue exposed by running task in parallel via nx
  -  now we use standard node_modules patching
- resolve `test` run issues
  - caused by invalid hoisting of npm workspaces  
- react targets are now run in parallel gaining 12% FASTER execution on CI

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -218,7 +218,7 @@ jobs:
     # svg-icons:build is run in different job (@see build-svg), -> exclude
     # svg-sprites:build doesn't work and is missing owner, -> exclude
       run: |
-        npx nx affected -t build --exclude='svg-icons,svg-sprites' --parallel 1
+        npx nx affected -t build --exclude='svg-icons,svg-sprites'
         npx nx affected -t build:verify --exclude='svg-icons,svg-sprites'
 
     - name: Test


dmytrokirpa (dismissed):

dmytrokirpa (dismissed):

packages/react-icons/vitest.config.ts
```diff
@@ -1,5 +1,23 @@
...
 export default defineConfig({
-  test: {},
+  test: {
+    globals: true,
+    setupFiles: ['./test-setup.js'],
+  },
+  plugins: [
+    {
+      name: 'mock-fonts',
+      resolveId(id) {
```
Hotell, line 11: makes vanilla `vitest` work withou need of `build` assets in place

package.json
```diff
@@ -12,10 +12,17 @@
...
   },
+  "scripts": {
+    "reset-workspace": "git clean -dXf",
+    "postinstall": "patch-package"
+  },
   "devDependencies": {
+    "eslint": "8.57.1",
     "nx": "21.6.5",
+    "patch-package": "8.0.1",
+    "react": "17.0.2",
```
Hotell, line 23: moved to root, vitest aliases overrides didn't work

packages/react-icons/build-verify.test.js
```diff
@@ -602,6 +602,7 @@ describe('Build Verification', () => {
         export * from './sizedIcons/chunk-34';
         export * from './sizedIcons/chunk-35';
         export * from './sizedIcons/chunk-36';
+        export * from './sizedIcons/chunk-37';
```
Hotell, line 605: updating snapshots from previous `/assets` changes not reflected in our toolchain https://github.com/microsoft/fluentui-system-icons/pull/903

packages/react-icons/src/utils/icon-factories.test.tsx
```diff
@@ -2,7 +2,7 @@ import * as React from 'react';
 import { render } from '@testing-library/react';
 import { describe, expect, test } from 'vitest';
 
-import { createFluentFontIcon, FluentFontIcon } from '../fonts';
+import { createFluentFontIcon, FluentFontIcon } from './fonts/createFluentFontIcon';
```
Hotell, line 5: avoids loading generated barrel

Issue #713, opened 2024-04-25
Add changelogs to repo


tomi-msft:
This issue is to track the work in adding changelogs to the repo. This is in order to keep track of changes being made to the repo in between version bumps. This particular issue is for for the `@fluentui/react-icons` package.

PR #908, opened 2025-10-20
chore: setup Nx Release for version and changelog generation for react-icons and hook it up to current publish flow

Hotell:
- sets Nx Release for `react-icons`
- generates initial `CHANGELOG.md`
- hooks up Nx Release to publish workflow in backwards-compatible way

**New Behaviour during release:**
- new variable `$CURRENT_VERSION` is created which is necessary for `nx release`
- react-icons `CHANGELOG.md` is updated based on changes since last release
- `package-lock.json` and test snapshots verifying chunks is updated during release to avoid creating invalid repo state

---

- Closes https://github.com/microsoft/fluentui-system-icons/issues/713

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -3,6 +3,7 @@
 /.github/react-icons.instructions.md @microsoft/cxe-prg
 /.github/workflows/copilot-setup-steps.yml @microsoft/cxe-prg
 /.github/workflows/pr.yml @microsoft/cxe-prg @spencer-nelson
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -17,6 +17,7 @@
     "postinstall": "patch-package"
   },
   "devDependencies": {
+    "@nx/js": "21.6.5",
```
Hotell, line 20: needed for node platform releases

PR #915, opened 2025-10-27
ci: implement pre-release publishing for npm/react-icons

Hotell:
Introduces a new workflow for publishing prerelease versions of the `@fluentui/react-icons` NPM package. 

This enables testing out new features in user land before hitting stable release.

**Automation of prerelease publishing:**

* Added a new GitHub Actions workflow, `publish-prerelease.yml`, to automate publishing prerelease versions of `@fluentui/react-icons` with support for `alpha`, `beta`, and `rc` tags. This workflow includes steps for building the package, calculating the next prerelease version, updating `package.json`, and publishing to NPM, with an option for dry runs.
* Introduced a new script, `.github/scripts/get-next-prerelease-version.js`, which determines the next prerelease version by querying the NPM registry and incrementing the version appropriately based on the latest published prerelease and the current `package.json`.

### DIFF
--- a/.github/scripts/get-next-prerelease-version.js
+++ b/.github/scripts/get-next-prerelease-version.js
@@ -6,18 +6,17 @@
  * Gets the next prerelease version for a package
  * by checking the latest version from npm registry with the specified preid
  *
- * Usage: node get-next-prerelease-version.js --package <package-name> --preid <preid> [--package-json <path>]
+ * Usage: node get-next-prerelease-version.js --package <package-name> --preid <preid>
  *
  * Options:
  *   --package: NPM package name (e.g., @fluentui/react-icons)
  *   --preid: Prerelease identifier (e.g., alpha, beta, rc)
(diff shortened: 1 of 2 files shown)


.github/scripts/get-next-prerelease-version.js
```diff
@@ -0,0 +1,268 @@
...
+ * by checking the latest version from npm registry with the specified preid
+ *
+ * Usage: node get-next-prerelease-version.js --package <package-name> --preid <preid>
+ *
+ * Options:
+ *   --package: NPM package name (e.g., @fluentui/react-icons)
+ *   --preid: Prerelease identifier (e.g., alpha, beta, rc)
+ */
+
+const fs = require('fs');
```
dmytrokirpa, line 16: nit: should these also be `node:fs` and `node:path`?

Hotell: yeah definitely `node:` prefix, will be normalized once we merge all the new features to mitigate churn

PR #930, opened 2025-11-17
Allow "Color" to be atomic icon name

### LINKED ISSUE #928: [Bug]: Color icons each get their own atomic js file
### Package version

2.0.314

### React version

18

### Environment

```shell
N/A
```

### Current Behavior

Icons like `Color16Filled` are exported from /lib/atoms/svg/color-16-filled.js

### Expected Behavior

Icons like `Color16Filled` should be exported from /lib/atoms/svg/color.js

### Reproduction

N/A

### Steps to reproduce

1. Build package
2. Look at atoms directory

### Are you reporting an Accessibility issue?

None

### Suggested severity

Medium - Has workaround

### Products/sites affected

_No response_

### Are you willing to submit a PR to fix?

yes

### Validations

- [x] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.
- [x] The provided reproduction is a minimal reproducible example of the bug.

CampbellOwen:
The current atomic icon file generation removes "color" in the `Color` icon as a style variant, leading to files like `color-16-filled.js` that should just be `color.js`.

This PR assumes the first "part" (separated by `-`) of the icon name is not a style variant and is actually the icon's name.

This will be a breaking change because all the `color-${size}-${variant}.js` files will be removed in favour of `color.js`

Fixes #928

### DIFF
--- a/packages/react-icons/per-icon.writer.js
+++ b/packages/react-icons/per-icon.writer.js
@@ -212,7 +212,7 @@ function normalizeBaseName(fileName, styleVariants = ICON_STYLE_VARIANTS) {
   const parts = normalized.split('-');
   const styleSet = new Set(styleVariants);
 
-  while (parts.length > 0) {
+  while (parts.length > 1) {
     const last = parts[parts.length - 1];
     // treat pure numbers and tokens that are in styleTokens or contain 'color' as variant markers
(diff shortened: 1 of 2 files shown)


Hotell:
I managed to implement it based on your suggestion in non breaking way, we can close this one https://github.com/microsoft/fluentui-system-icons/pull/943

ty @CampbellOwen !

PR #953, opened 2026-01-20
feat(react-icons): introduce granular export maps for svg utils

Hotell:
Introduces another public api endpoint for svg related utilities to improve tree-shaking and bundle size -> `@fluentui/react-icons/utils`

**Before:**
```ts
import {AccessTimeFilled, AccessTimeRegular} from '@fluentui/react-icons/svg/access-time';
import {bundleIcon, iconRegularClassName} from '@fluentui/react-icons';

const Bundled = bundleIcon(AccessTimeFilled,AccessTimeRegular);
```

**After**

```ts
import {AccessTimeFilled, AccessTimeRegular} from '@fluentui/react-icons/svg/access-time'
// 🚢 NEW api
import {bundleIcon, iconRegularClassName} from '@fluentui/react-icons/utils';

const Bundled = bundleIcon(AccessTimeFilled,AccessTimeRegular);
```

### Additional changes

- changes `bundleIcon` from default to named export for better three shaking
- fixes invalid custom `filled` property propagation to `<svg>`

### DIFF
--- a/packages/react-icons/README.md
+++ b/packages/react-icons/README.md
@@ -227,6 +227,21 @@ module.exports = {
       {
         '@fluentui/react-icons': {
           transform: (importName) => {
+            // Handle utility imports (bundleIcon, className constants)
+            const utilityExports = [
+              'bundleIcon',
+              'iconClassName',
+              'iconFilledClassName',
+              'iconRegularClassName',
+              'iconColorClassName',
+              'iconLightClassName'
+            ];
+
+            if (utilityExports.includes(importName)) {
+              return '@fluentui/react-icons/utils';
+            }
+
+            // Handle icon imports
             const withoutSuffix = importName.replace(
               /(\d*)?(Regular|Filled|Light|Color)$/,
               ''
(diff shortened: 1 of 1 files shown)


packages/react-icons/src/utils/bundleIcon.tsx
```diff
@@ -6,35 +6,23 @@ import type { FluentIcon } from "./createFluentIcon";
...
-                />
-            </React.Fragment>
-        )
-    }
-    Component.displayName = "CompoundIcon";
-    return Component;
-}
-
-export default bundleIcon;
+export const bundleIcon = (FilledIcon: FluentIcon, RegularIcon: FluentIcon): FluentIcon => {
```
Hotell, line 9: added only explicit return annotation - the body was reformatted by prettier

packages/react-icons/src/utils/FluentIconsProps.types.ts
```diff
@@ -4,11 +4,15 @@ export type FluentIconsProps<
...
   React.RefAttributes<TRefType> & {
     primaryFill?: string;
     className?: string;
-    filled?: boolean;
     title?: string;
+    /**
+     * Whether to render the filled version of the icon.
+     * NOTE: This prop is only applicable when using bundled icons created with `bundleIcon`.
+     */
+    filled?: boolean;
```
dmytrokirpa, line 17: Does it make sense to have two separate types like:

```ts
export type FluentIconsProps<
  TBaseAttributes extends
    | React.SVGAttributes<SVGElement>
    | React.HTMLAttributes<HTMLElement> = React.SVGAttributes<SVGElement>,
  TRefType extends HTMLElement | SVGSVGElement = SVGSVGElement,
> = TBaseAttributes &
  React.RefAttributes<TRefType> & {
    primaryFill?: string;
    className?: string;
    title?: string;
}

export type FluentCompoundIconProps<
  TBaseAttributes extends
    | React.SVGAttributes<SVGElement>
    | React.HTMLAttributes<HTMLElement> = React.SVGAttributes<SVGElement>,
  TRefType extends HTMLElement | SVGSVGElement = SVGSVGElement,
> = FluentIconsProps<TBaseAttributes> & {
/**
     * Whether to render the filled version of the icon.
     */
    filled?: boolean;
}
```

Hotell: it does, but if we do this it would break user pipeline if someone "by accident" set `filled` on normal icon

PR #955, opened 2026-01-22
fix(react-icons): properly resolve atoms grouppings if icon contains style variant as part of its name

Hotell:
**Before:**

`TextColor<size><variant>` exports have been incorrectly groupped with Text<size><variant> under `text.tsx` module

> source: https://app.unpkg.com/@fluentui/react-icons@2.0.317/files/lib/atoms/svg/text.js
>  issue showcase

**After:**

- `TextColor<size><variants>` exports groupped under `text-color.tsx`
- `Text<size><variant>` exports groupped within `text.tsx` module
  - to mitigate any breaking changes we add the TextColor* export back with `@deprecation` annotation
  - image

This fix enables build transforms via swc or babel without additional workarounds as initially documented  https://github.com/microsoft/fluentui-system-icons/blob/main/packages/react-icons/README.md#using-api-via-build-transform

---

- Follows https://github.com/microsoft/fluentui-system-icons/pull/943

### DIFF
--- a/packages/react-icons/build-verify.test.js
+++ b/packages/react-icons/build-verify.test.js
@@ -1975,9 +1975,9 @@ describe('Build Verification', () => {
         expect(textColorContent).not.toContain('export const Text16Regular');
 
         // text.js should contain Text* exports
-        expect(textContent).toMatch(/export const Text12Regular/);
-        expect(textContent).toMatch(/export const Text16Regular/);
-        expect(textContent).toMatch(/export const Text32Regular/);
+        expect(textContent).toContain('export const Text12Regular');
+        expect(textContent).toContain('export const Text16Regular');
+        expect(textContent).toContain('export const Text32Regular');
 
         // text.js should have backward-compatible re-exports for TextColor* with deprecation notice
         expect(textContent).toContain(`@deprecated use \`@fluentui/${exportKindDir}/text-color\` import`);


packages/react-icons/per-icon.writer.js
```diff
@@ -197,45 +197,58 @@ async function writePerIconFiles(destPath, items, headerLines = [], options = {
...
+ * @example
+ *  - 'zoom-in-20-filled.tsx' -> 'zoom-in'
+ *  - 'my-icon-16-regular' -> 'my-icon'
+ *
+ * @param {string} fileName - The raw file name or path component to normalize.
+ * @param {string[]} [styleVariants=ICON_STYLE_VARIANTS] - The list of allowed style suffixes to identify and remove.
+ * @returns {string} The normalized base name of the icon.
+ * @throws {Error} If the filename does not end with a valid style suffix.
+ */
+function normalizeBaseName(fileName, styleVariants = ICON_STYLE_VARIANTS) {
```
Hotell, line 234: implementation was simplified with proper guards to adhere to valid file patterns from `/assets` provided by design

packages/react-icons/per-icon.writer.test.js
```diff
@@ -41,9 +41,13 @@ describe('writePerIconFiles', () => {
     fs.mkdirSync(tmpDest, { recursive: true });
     const items = [
       { exportName: 'Dup20Filled', exportCode: 'export const Dup20Filled = 1;', fileName: 'dup-20-filled.tsx' },
-      { exportName: 'Dup20Filled', exportCode: 'export const Dup20Filled = 2;', fileName: 'dup-20-filled-2.tsx' },
+      { exportName: 'Dup20Filled', exportCode: 'export const Dup20Filled = 2;', fileName: 'dup-22-filled.tsx' },
```
Hotell, line 44: we had "made up" patterns within test which doesn't exist, aligned now with real contents of `/assets`

packages/react-icons/deprecated-atoms.js
```diff
@@ -52,4 +85,164 @@ function handleDeprecatedColorAtoms(destPath, type) {
...
+ * Style variants: regular, filled, light, color
+ * Potential problem patterns:
+ *   - Icon names like "Something Light", "Border Regular", etc.
+ *   - Would normalize to the same base name as other icons without the variant suffix
+ *
+ * @param {string} destPath - The destination path where atoms are written.
+ * @returns {Promise<void>}
+ * @throws Will throw an error if potential grouping issues are detected.
+ */
+async function assertCompoundStyleVariantIssues(destPath) {
```
Hotell, line 138: added build last line of defense to double check generated outputs from `/assets`, will fail build if there are issues

packages/react-icons/deprecated-atoms.js
```diff
@@ -22,14 +56,13 @@ function handleDeprecatedColorAtoms(destPath, type) {
     return;
   }
 
-  const deprecatedMsg = `/** @deprecated use from @fluentui/${type}/color import. This was generated for backward compatibility and will be removed in next major release */`;
-  const generatedFiles = []
+  const deprecatedMsg = getDeprecatedMsg(type, 'color');
```
Hotell, line 59: DRY applied

packages/react-icons/convert.utils.test.js
```diff
@@ -112,9 +112,9 @@ describe(`convert  utils`, () => {
           if (fs.existsSync(tmpDest)) fs.rmSync(tmpDest, { recursive: true, force: true });
 
           // Two different filenames that will produce the same export name when processed
-          writeSvg('ic_fluent_dup_20.svg', '<svg width="20" d="M0"></svg>');
-          // this variant without underscore will camelCase to the same export name (Dup20)
-          writeSvg('ic_fluent_dup20.svg', '<svg width="20" d="M1"></svg>');
+          writeSvg('ic_fluent_dup_20_regular.svg', '<svg width="20" d="M0"></svg>');
```
Hotell, line 154: updates test to reflect reality of `/assets`

packages/react-icons/per-icon.writer.js
```diff
@@ -197,45 +197,58 @@ async function writePerIconFiles(destPath, items, headerLines = [], options = {
...
- * @param {string[]=} styleVariants
+ * @param {string[]} [styleVariants]
+ * @returns
  */
-function normalizeBaseName(fileName, styleVariants = ICON_STYLE_VARIANTS) {
-  const name = fileName.replace(/\.tsx?$/, '');
-  // normalize separators (underscores -> hyphens) then split
-  const normalized = name.replace(/_/g, '-');
+function parseIconFileName(fileName, styleVariants = ICON_STYLE_VARIANTS) {
+  const nameWithoutFileExt = fileName.replace(/\.tsx?$/, '');
```
mainframev, line 208: can't `path.parse(fileName).name` be used here?

packages/react-icons/deprecated-atoms.test.js
```diff
@@ -0,0 +1,136 @@
+// @ts-check
+import fs from 'fs';
+import path from 'path';
+import { describe, it, expect, beforeEach, afterEach } from 'vitest';
+import { handleDeprecatedColorAtoms, handleDeprecatedTextColorAtoms } from './deprecated-atoms';
+
+describe(`deprecated-atoms`, () => {
+  const tmpDest = path.join(__dirname, 'tmp-deprecated-atoms-test');
```
mainframev, line 8: won't be better here to use `fs.mkdtempSync` and `os.tmpdir` ?

mainframev (dismissed):

packages/react-icons/deprecated-atoms.test.js
Hotell: absolutely, keeping consistent with existing tests though, we can refactor as batch later

Hotell: https://github.com/microsoft/fluentui-system-icons/issues/971

packages/react-icons/per-icon.writer.js
Hotell: it can, question is if we need a node `path` library for this ? having that would be 100% more generic, but that's not necessary in our case atm

mainframev (approved):

PR #961, opened 2026-01-23
feat(react-icons-font-subsetting-webpack-plugin): add react-icons atomic imports support

Hotell:
- [x] needs https://github.com/microsoft/fluentui-system-icons/pull/960

---

Adds support for font subsetting when using atomic imports  `@fluentui/react-icons/fonts/*`

### DIFF
--- a/packages/react-icons-font-subsetting-webpack-plugin/src/index.ts
+++ b/packages/react-icons-font-subsetting-webpack-plugin/src/index.ts
@@ -106,7 +106,10 @@ function isNormalModule(m: webpack.Module): m is webpack.NormalModule {
 
 function isFluentUIReactFontChunk(m: webpack.Module): m is webpack.NormalModule {
     if (isNormalModule(m)) {
-        return /react-icons[\/\\]lib(-cjs)?[\/\\]fonts[\/\\](sizedIcons|icons)[\/\\]chunk-\d+\.js$/.test(m.resource)
+        // Match both chunk files and atomic font imports:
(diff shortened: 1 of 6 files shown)


packages/react-icons-font-subsetting-webpack-plugin/src/index.ts
```diff
@@ -106,7 +106,10 @@ function isNormalModule(m: webpack.Module): m is webpack.NormalModule {
 
 function isFluentUIReactFontChunk(m: webpack.Module): m is webpack.NormalModule {
     if (isNormalModule(m)) {
-        return /react-icons[\/\\]lib(-cjs)?[\/\\]fonts[\/\\](sizedIcons|icons)[\/\\]chunk-\d+\.js$/.test(m.resource)
+        // Match both chunk files and atomic font imports:
+        // - lib/fonts/sizedIcons/chunk-0.js (chunk-based)
+        // - lib/atoms/fonts/access-time.js (atomic imports)
+        return /react-icons[\/\\]lib(-cjs)?[\/\\](fonts[\/\\](sizedIcons|icons)[\/\\]chunk-\d+|atoms[\/\\]fonts[\/\\][\w-]+)\.js$/.test(m.resource)
```
dmytrokirpa, line 112: not sure how often this fn is called, but could we move the RegExp to a module scope and reuse the instance?

Hotell: yup I was thinking the same, but decided to keep the feature scoped, will address as follow up

PR #968, opened 2026-02-02
ci(publish): extend repo checkout to last 50 commits to provide necessary tags and history for Nx release

Hotell:
- release fails as Nx Release is missing tags and commit history which are used for changelog generation / https://github.com/microsoft/fluentui-system-icons/actions/runs/21527610253/job/62035201445

This PR
-  sets checkout history for 50 last commits
- updates nx release configuration to mitigate bumping workspace dependency ranges for `devDependencies` ( this mitigates changing `*` version specifiers )

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -36,10 +36,11 @@ jobs:
     - uses: actions/checkout@v6
       with:
         token: ${{ steps.generate-token.outputs.token }}
-        # This is needed for nx release changelog to detect all commits between releases
+        #  Git knows file names/paths for history, but blobs (file contents) are fetched on demand.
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -162,16 +186,15 @@ jobs:
...
-      run: |
-        npm run build
-      working-directory: packages/react-icons-font-subsetting-webpack-plugin
-
     - name: Build React-Native library
       run: |
         npm run build
       working-directory: packages/react-native-icons
 
+    - name: Build Webpack plugin library
```
Hotell, line 194: this is onboarded to nx so no need to run from different working directory

PR #969, opened 2026-02-02
ci(cocoapods-publish): allow manual release trigger and schedule release from publish library workflow

Hotell:
cocoapods releases are no longer triggered after https://github.com/microsoft/fluentui-system-icons/pull/962 - this works as expected because the "skip ci" commit suffix

**New behaviour:**

- manual workflow works
- during publish, cocoapods publish is triggered ( if not in dry-run mode )

### DIFF
--- a/.github/workflows/cocoapods-publish.yml
+++ b/.github/workflows/cocoapods-publish.yml
@@ -4,7 +4,6 @@ on:
   workflow_dispatch:
 
 jobs:
-  # Runs after the main publish workflow completes successfully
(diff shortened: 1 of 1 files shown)


.github/workflows/cocoapods-publish.yml
```diff
@@ -2,14 +2,11 @@ name: Cocoapods publish
 
 on:
   workflow_dispatch:
-  push:
-    branches: [ main ]
 
 jobs:
-  # Only runs on the version update commit
-  # We need the new podspec available to publish to CocoaPods
+  # Runs after the main publish workflow completes successfully
```
Hotell:
```suggestion
```

PR #1010, opened 2026-03-18
feat(react-icons): deprecate colored svg variants

ValentinaKozlova:
Adds /** @deprecated */ JSDoc comment to all Color style icon exports in the SVG generation pipeline
Color icons are not recommended due to accessibility concerns (not HCM compliant, contrast issues, SVG namespace conflicts)
The deprecation tag will surface warnings in IDEs/editors when consumers import Color variants, guiding them toward Filled or Regular alternatives


packages/react-icons/convert.utils.js
```diff
@@ -56,13 +56,17 @@ function makeIconExport(opts) {
           ? ', { color: true }'
           : '';
 
+  const deprecatedPrefix = isColor
```
Hotell: this covers only chunks, but needs to be in all api ( including atoms )

https://github.com/microsoft/fluentui-system-icons/blob/main/packages/react-icons/convert.utils.js#L65

packages/react-icons/build-verify.test.js
```diff
@@ -1329,6 +1329,7 @@ describe('Build Verification', () => {
           export declare const CalculatorArrowClockwiseRegular: FluentIcon;
           export declare const CalculatorMultipleFilled: FluentIcon;
           export declare const CalculatorMultipleRegular: FluentIcon;
+          /** @deprecated Color icons are not recommended due to accessibility concerns (not HCM compliant, contrast issues). Consider using Filled or Regular variants instead. */
```
Hotell: this deprecation message isn't that helpful unfortunately. we need to create docs which will provide guidance to agents/devs on how to migrate including examples. then we can provide the link to the deprecation message. not blocking for this PR

packages/react-icons/convert.utils.js
Hotell: the deprecated annotation needs to be added to:
- chunk generation
- atomic svgs -> https://github.com/microsoft/fluentui-system-icons/blob/main/packages/react-icons/convert.utils.js#L138
- atomic svg sprites -> https://github.com/microsoft/fluentui-system-icons/blob/main/packages/react-icons/sprite.writer.js
- all `/base` apis -> https://github.com/microsoft/fluentui-system-icons/blob/main/packages/react-icons/convert.utils.js#L138

packages/react-icons/build-verify.test.js
Hotell: we have now SB deployed, lets use that for reference link

packages/react-icons/build-verify.test.js
Hotell, line 1: lets update svg atoms test to cover that those have deprecation as well please

Hotell (approved):
build-verify test need some extended coverage - please see comments. otherwise LGTM

packages/react-icons/metadata.json
```diff
@@ -121169,6 +121389,16 @@
     "font": true,
     "type": "sized"
   },
+  "TextUnderline28Filled": {
```
Hotell, line 121392: these should have been added by recent release, have you updated your branch against latest main ?

ValentinaKozlova: yes

PR #1048, opened 2026-04-02
feat(react-icons-atomic-webpack-loader): implement build-transform webpack loader

mainframev:
The idea of adding that plugin is a result of explorations in user land code with adding build transforms to support react-icons atomic API. 

This should help users migrate to atomic API in a more convenient way. 

The loader respects atomic API routing for svg / font icons and applies on both imports and reexports.

### DIFF
--- a/nx.json
+++ b/nx.json
@@ -16,6 +16,7 @@
       "react": {
         "projects": [
           "packages/react-icons",
+          "packages/react-icons-atomic-webpack-loader",
           "packages/react-icons-font-subsetting-webpack-plugin",
           "packages/react-native-icons"
         ],
--- a/nx.json
+++ b/nx.json
@@ -16,7 +16,6 @@
       "react": {
         "projects": [
           "packages/react-icons",
-          "packages/react-icons-atomic-webpack-loader",
           "packages/react-icons-font-subsetting-webpack-plugin",
           "packages/react-native-icons"
         ],
--- a/packages/react-icons-atomic-webpack-loader/README.md
+++ b/packages/react-icons-atomic-webpack-loader/README.md
@@ -1,6 +1,9 @@
 # @fluentui/react-icons-atomic-webpack-loader
 
-> ⚠️ This package is in early development. The API may change before the first stable release.
+> **⚠️ Alpha** — this package is available as an alpha prerelease only.
+> ⚠️ The API may change before the first stable release.
+
+> Install via `npm install @fluentui/react-icons-atomic-webpack-loader@alpha --save-dev`
 
 Webpack loader that transforms barrel imports and re-exports from `@fluentui/react-icons` into atomic deep paths for better tree-shaking and smaller bundles.
 
@@ -18,15 +21,11 @@ import { useIconContext } from '@fluentui/react-icons/providers';
 export { ArrowLeftRegular } from '@fluentui/react-icons/svg/arrow-left';
 ```
 
-## Install
-
-```bash
-npm install @fluentui/react-icons-atomic-webpack-loader --save-dev
-```
-
 ## Usage
 
-Add the loader to your webpack config. It should run on both your source files and any `node_modules` that re-export from `@fluentui/react-icons`.
+Add the loader to your webpack config as an [`enforce: 'pre'`](https://webpack.js.org/configuration/module/#ruleenforce) rule so it runs on the original source before any other loaders:
+
+NOTE: Unlike most loaders, this one should NOT exclude `node_modules`. It needs to process files inside `node_modules` as well to transform barrel imports from `@fluentui/react-icons` in your third-party dependencies. Files that don't reference `@fluentui/react-icons` are skipped via a fast regex pre-check, so there is no meaningful overhead.
 
 ```js
 // webpack.config.js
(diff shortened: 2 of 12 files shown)


nx.json
```diff
@@ -16,6 +16,7 @@
       "react": {
         "projects": [
           "packages/react-icons",
+          "packages/react-icons-atomic-webpack-loader",
```
Hotell: lets remove this, we will release this in prerelease tag from branch for now. ATM the stable release process uses same semver for every package, which is something we wanna avoid for this. we need to come up with a more flexible release flow/process from main.

packages/react-icons-atomic-webpack-loader/package.json
```diff
@@ -0,0 +1,37 @@
+{
+  "name": "@fluentui/react-icons-atomic-webpack-loader",
+  "version": "0.1.0",
```
Hotell: needs to be private to avoid release with stable

```suggestion
  "version": "0.0.0",
  "private": true
```

packages/react-icons-atomic-webpack-loader/package.json
```diff
@@ -0,0 +1,37 @@
...
+  "dependencies": {
+    "acorn-typescript": "^1.4.13"
+  },
+  "devDependencies": {
+    "typescript": "5.0.4",
+    "webpack": "^5.72.0",
+    "@types/node": "14"
+  },
+  "peerDependencies": {
+    "acorn": ">=8.9.0",
```
Hotell, line 34: is acorn part of webpack ? if not we should add it to deps similarily to acorn ts plugin

packages/react-icons-atomic-webpack-loader/test/webpack.config.js
```diff
@@ -0,0 +1,104 @@
...
+    context: __dirname,
+    mode: 'production',
+    optimization: { minimize: false },
+    entry: { [name]: entry.src },
+    output: {
+      path: resolve(__dirname, 'dist', name),
+      filename: '[name].js',
+      libraryTarget: 'commonjs2',
+    },
+    externals: [/^@fluentui\/react-icons/, /^react$/, /^other-package$/],
```
Hotell: what is `other-package` for ?

packages/react-icons-atomic-webpack-loader/README.md
```diff
@@ -0,0 +1,90 @@
+# @fluentui/react-icons-atomic-webpack-loader
+
+> ⚠️ This package is in early development. The API may change before the first stable release.
```
Hotell: lets follow our pattern https://github.com/microsoft/fluentui-system-icons/blob/90239bc59dc89b20ca27d74dac74c93a410a9d8a/packages/react-icons-svg-sprite-subsetting-webpack-plugin/README.md?plain=1#L3-L5

packages/react-icons-atomic-webpack-loader/test/webpack.config.js
```diff
@@ -0,0 +1,104 @@
...
+  return {
+    name,
+    context: __dirname,
+    mode: 'production',
+    optimization: { minimize: false },
+    entry: { [name]: entry.src },
+    output: {
+      path: resolve(__dirname, 'dist', name),
+      filename: '[name].js',
+      libraryTarget: 'commonjs2',
```
Hotell: commonjs ? lets keep it esm

packages/react-icons-atomic-webpack-loader/README.md
```diff
@@ -0,0 +1,90 @@
...
+// webpack.config.js
+module.exports = {
+  module: {
+    rules: [
+      {
+        test: /\.[mc]?[jt]sx?$/,
+        exclude: /node_modules/,
+        use: [
+          'your-existing-loader', // e.g. babel-loader, ts-loader, swc-loader
+          '@fluentui/react-icons-atomic-webpack-loader',
```
Hotell: shouldn't this be rather via `enforce:pre` https://webpack.js.org/configuration/module/#ruleenforce ?

packages/react-icons-atomic-webpack-loader/package.json
```diff
@@ -0,0 +1,37 @@
...
+  "bugs": {
+    "url": "https://github.com/microsoft/fluentui-system-icons/issues"
+  },
+  "dependencies": {
+    "acorn-typescript": "^1.4.13"
+  },
+  "devDependencies": {
+    "typescript": "5.0.4",
+    "webpack": "^5.72.0",
+    "@types/node": "14"
```
Hotell: lets use similar config as in our latest loader https://github.com/microsoft/fluentui-system-icons/blob/main/packages/react-icons-svg-sprite-subsetting-webpack-plugin/package.json#L36

Hotell (commented):
left some comments, looks good ! ty

packages/react-icons-atomic-webpack-loader/package.json
mainframev: yes, that's why I added that to peer deps

packages/react-icons-atomic-webpack-loader/src/transform.ts
```diff
@@ -0,0 +1,127 @@
...
+        lines.push(`export { ${spec} } from '${newSource}';`);
+      }
+
+      replacements.push({ start: n.start, end: n.end, text: lines.join('\n') });
+    }
+  }
+
+  // Apply in reverse to preserve character offsets
+  let result = source;
+  for (let i = replacements.length - 1; i >= 0; i--) {
```
Hotell: we should probably use https://github.com/Rich-Harris/magic-string to preserve formatting and make it faster ?

mainframev: Good suggestions, ty!

packages/react-icons-atomic-webpack-loader/test/webpack.config.js
mainframev: because of the fixture :) removed

packages/react-icons-atomic-webpack-loader/src/transform.ts
```diff
@@ -0,0 +1,129 @@
...
+export function transformSource(source: string, options: TransformOptions): TransformResult {
+  const parser = getParser(options);
+
+  const ast = parser.parse(source, {
+    sourceType: 'module',
+    ecmaVersion: 'latest',
+    locations: false,
+  });
+
+  const s = new MagicString(source);
```
Hotell: lets use more human readable naming if possible

Hotell (approved):
one nit, LGTM, ty

packages/react-icons-atomic-webpack-loader/README.md
```diff
@@ -0,0 +1,84 @@
...
+
+Add the loader to your webpack config as an [`enforce: 'pre'`](https://webpack.js.org/configuration/module/#ruleenforce) rule so it runs on the original source before any other loaders:
+
+```js
+// webpack.config.js
+module.exports = {
+  module: {
+    rules: [
+      {
+        test: /\.[mc]?[jt]sx?$/,
```
Hotell, line 36: we should also add "NOTE" that this needs to run on installed packages in node_modules that use/re-export react-icons

PR #1052, opened 2026-04-10
docs: build-transforms.md to include webpack loader

mainframev:
Updates readme to mention `react-icons-atomic-webpack-loader`  with atomic API


Hotell (approved):
LGTM

PR #1061, opened 2026-04-14
feat(react-icons-atomic-webpack-loader): replace acorn parser with oxc-parser

mainframev:
After first alpha release user reported few issues with acorn and typescript. 

~~Original `acorn-typescript` is no longer maintained and does not support ts features above 4.9. There is a svelte fork that is still being maintained. Added a regex based fallback, so if acorn fails for some reason to parse the file content, it won't silently skip the transform~~

Replaces parser from Acorn to `oxc-parser`, simplifying loader logic, and adding testing for both `.jsx` and `.tsx` files.


packages/react-icons-atomic-webpack-loader/test/src/ts-satisfies.ts
```diff
@@ -0,0 +1,6 @@
+import { AddFilled } from '@fluentui/react-icons';
```
mainframev: makes sure that new acorn-typescript fork supports TS > 4.9 features like `satisfies`. Original `acorn-typescript` fails to parse files with that syntax

packages/react-icons-atomic-webpack-loader/src/index.ts
```diff
@@ -11,21 +12,32 @@ export default function fluentIconsAtomicImportLoader(
...
-    const { code, map } = transformSource(sourceCode, { iconVariant, isTypescript, isTsx });
-    this.callback(null, code, map);
+    result = transformSource(sourceCode, { iconVariant, isTypescript, isTsx });
   } catch {
-    this.emitWarning(new Error(`FluentIconsAtomicImportLoader: Failed to transform "${this.resourcePath}"`));
-    this.callback(null, sourceCode);
+    result = null;
   }
+
+  if (!result) {
```
mainframev: A regex-based fallback that rewrites imports when acorn meets unsupported syntax

packages/react-icons-atomic-webpack-loader/test/src/tsx-spread-children.tsx
```diff
@@ -0,0 +1,6 @@
+import { AddFilled } from '@fluentui/react-icons';
+
+const items = [<div key="a" />];
+const el = <div>{...items}</div>;
```
mainframev: this is added for scenario, when acorn meets unsupported syntax

packages/react-icons-atomic-webpack-loader/package.json
```diff
@@ -3,10 +3,11 @@
   "version": "0.0.0",
   "private": true,
   "description": "Webpack loader that transforms barrel imports and re-exports from @fluentui/react-icons into atomic deep paths",
+  "type": "module",
```
mainframev: `acorn-typescript` is ESM-only package

packages/react-icons-atomic-webpack-loader/src/index.ts
Hotell: good idea with this one, but shouldn't the forked acorn cover all the syntax ?

packages/react-icons-atomic-webpack-loader/test/esbuild-loader.js
```diff
@@ -0,0 +1,11 @@
+// @ts-check
+const { transformSync } = require('esbuild');
+
+/** @type {import('webpack').LoaderDefinitionFunction} */
+module.exports = function (source) {
```
Hotell: cant we use what we already have in repo ? ts-loader

packages/react-icons-atomic-webpack-loader/test/src/tsx-spread-children.tsx
Hotell: what's the unsupported syntax in this case ? spread ?

packages/react-icons-atomic-webpack-loader/test/src/ts-same-name.ts
```diff
@@ -0,0 +1,6 @@
+import { AddFilled } from '@fluentui/react-icons';
+
+type Foo = { bar: string };
```
Hotell: acorn is unable to parse file which exposes merged symbol ? ( type + runtime var ) , this has been in TS since version 2.0 AFAIR, is this known acorn issue?

packages/react-icons-atomic-webpack-loader/src/transform.ts
```diff
@@ -49,13 +49,58 @@ export interface TransformResult {
...
+  /(import(?:\s+type)?|export(?:\s+type)?)\s*\{([^}]+)\}\s*from\s*['"]@fluentui\/react-icons['"]\s*;?/g;
+
+const SPECIFIER_RE = /(?:type\s+)?(\w+)(?:\s+as\s+(\w+))?/g;
+
+export function transformSourceRegex(source: string, options: Pick<TransformOptions, 'iconVariant'>): TransformResult {
+  const src = new MagicString(source);
+  let match: RegExpExecArray | null;
+
+  IMPORT_EXPORT_RE.lastIndex = 0;
+  while ((match = IMPORT_EXPORT_RE.exec(source)) !== null) {
```
Hotell: im worried about performance of this fallback approach. if the forked acorn doesn't support full ts syntax we need to replace it. for sake of getting users unblock we can proceed with this one, but we should optimize if possible

packages/react-icons-atomic-webpack-loader/package.json
Hotell: will this force users to use mjs for webpack configuration ?

Hotell (commented):
left some comments, looks good so far ty

packages/react-icons-atomic-webpack-loader/src/index.ts
mainframev: I think it should, however it seems with the spread syntax it was not covered. Sorry for the mess, but I've decided to move to oxc-parser as we discussed in DMs. I believe that's more "bulletproof" way and won't bring some surprises in futures, as it has native built-in ts support 🙂

packages/react-icons-atomic-webpack-loader/test/src/tsx-spread-children.tsx
mainframev: yes

packages/react-icons-atomic-webpack-loader/package.json
mainframev: from what I've check it should support it without explicit mjs extension for config, after 5.80 version

packages/react-icons-atomic-webpack-loader/src/transform.ts
```diff
@@ -1,17 +1,14 @@
-import * as acorn from 'acorn';
-import { tsPlugin } from 'acorn-typescript';
+import { parseSync } from 'oxc-parser';
 import MagicString from 'magic-string';
 import type { ImportDeclaration, ExportNamedDeclaration, ImportSpecifier, ExportSpecifier } from 'estree';
```
layershifter: OXC exports own types, they are not 1/1 to `estree`. Plz switch to them

packages/react-icons-atomic-webpack-loader/src/transform.ts
```diff
@@ -35,29 +32,24 @@ function getName(node: { type: string; name?: string; value?: unknown }): string
...
   });
 
+  if (result.errors.length > 0) {
+    throw new Error(result.errors[0].message);
+  }
+
+  const ast = result.program;
   const src = new MagicString(source);
 
   for (const node of ast.body) {
```
layershifter: You don't need to do any AST work as OXC provides on `result` `staticImports`:

Image

(it includes info about specifiers & locs, so it's enough to perform modifications)

Here is a relevant part from Griffel transformer: https://github.com/microsoft/griffel/blob/93ef3aebe23ada447170c3846bd92d7e52f21573/packages/transform/src/transformSync.mts#L214

packages/react-icons-atomic-webpack-loader/src/transform.ts
```diff
@@ -35,29 +32,24 @@ function getName(node: { type: string; name?: string; value?: unknown }): string
...
 export function transformSource(source: string, options: TransformOptions): TransformResult {
-  const parser = getParser(options);
+  const { iconVariant, isTypescript, path } = options;
 
-  const ast = parser.parse(source, {
+  const result = parseSync(path, source, {
     sourceType: 'module',
-    ecmaVersion: 'latest',
-    locations: true,
+    astType: isTypescript ? 'ts' : 'js',
```
layershifter: AFAIR you don't need to pass it explicit, `parseSync()` has auto detection by path

packages/react-icons-atomic-webpack-loader/src/index.ts
```diff
@@ -11,21 +12,20 @@ export default function fluentIconsAtomicImportLoader(
   this: LoaderContext<FluentIconsAtomicImportLoaderOptions>,
   sourceCode: string,
 ): void {
+  const { resourcePath } = this;
+
   if (!REACT_ICONS_IMPORT_REGEX.test(sourceCode)) {
```
layershifter: Is regex needed there? Bailout should be cheap, so `indexOf` / `includes` should be enough

packages/react-icons-atomic-webpack-loader/src/index.ts
```diff
@@ -11,21 +12,20 @@ export default function fluentIconsAtomicImportLoader(
...
-    return;
+    return this.callback(null, sourceCode);
   }
 
-  const options = this.getOptions();
-  const iconVariant = options.iconVariant ?? 'svg';
-  const isTsx = this.resourcePath.endsWith('.tsx');
-  const isTypescript = /\.[mc]?tsx?$/.test(this.resourcePath);
+  const { iconVariant = 'svg' } = this.getOptions();
+  const isTypescript = TS_EXT_REGEX.test(resourcePath);
```
layershifter: Try to avoid regexes in hotpaths. Also `isTypescript` might be redundant, check other comment

packages/react-icons-atomic-webpack-loader/src/index.ts
```diff
@@ -11,21 +12,20 @@ export default function fluentIconsAtomicImportLoader(
...
   try {
-    const { code, map } = transformSource(sourceCode, { iconVariant, isTypescript, isTsx });
-    this.callback(null, code, map);
+    const { code, map } = transformSource(sourceCode, { iconVariant, isTypescript, path: resourcePath });
+    return this.callback(null, code, map);
   } catch {
-    this.emitWarning(new Error(`FluentIconsAtomicImportLoader: Failed to transform "${this.resourcePath}"`));
-    this.callback(null, sourceCode);
+    this.emitWarning(new Error(`FluentIconsAtomicImportLoader: Failed to transform "${resourcePath}"`));
+    return this.callback(null, sourceCode);
```
layershifter:
```suggestion
    return this.callback(new Error(`FluentIconsAtomicImportLoader: Failed to transform "${resourcePath}"`));
```

Don't you want to fail build actually?

packages/react-icons-atomic-webpack-loader/src/transform.ts
```diff
@@ -35,29 +32,24 @@ function getName(node: { type: string; name?: string; value?: unknown }): string
...
   map: ReturnType<MagicString['generateMap']>;
 }
 
 export function transformSource(source: string, options: TransformOptions): TransformResult {
-  const parser = getParser(options);
+  const { iconVariant, isTypescript, path } = options;
 
-  const ast = parser.parse(source, {
+  const result = parseSync(path, source, {
     sourceType: 'module',
```
layershifter: I would remove `sourceType` and let it detect, but it's reasonable to assume that we deal with ESM

Hotell (approved):
LGTM

layershifter (approved):

PR #1062, opened 2026-04-16
ci: support standalone project releases via separate release group and pipeline

Hotell:
> Needs https://github.com/microsoft/fluentui-system-icons/pull/1063

## Summary

Adds a new `standalone` release group and a dedicated publish workflow for independently-versioned npm packages that don't follow the main library's single-version policy.

The main goal behind this is to avoid "pre-release" publishing for packages that didnt have initial official version. Also this aligns with fluent publishing process where we embrace **zero major version** versioning by default when shipping new chunk of functionality (package) while validating it's stability against our user base.

### Changes

- **`nx.json`** — Added `standalone` release group with `projectsRelationship: "independent"` and `adjustSemverBumpsForZeroMajorVersion: true` (pending nx upgrade)
- **`packages/react-icons-atomic-webpack-loader`** — Removed `private: true`, added `npm:standalone` tag
- **`packages/react-icons-svg-sprite-subsetting-webpack-plugin`** — Removed `private: true`, added `npm:standalone` tag
- **`.github/workflows/publish-standalone.yml`** — New workflow:
  - Manually triggered from `main` only, with dry-run support
  - Resolves projects dynamically via `nx show projects -p "tag:npm:standalone"`
  - Uses conventional commits to determine version bumps (`nx release version`)
  - Generates per-project changelogs (`nx release changelog`)
  - Publishes via `nx release publish`
  - Creates per-package git tags (`{projectName}@{version}`) after commit for changelog range tracking
- **`docs/releases.md`** — Documented the standalone release group and new workflow

### How it works

- Version bumps are driven by conventional commits (inherited from shared nx config)
- Per-package git tags (`react-icons-atomic-webpack-loader@0.0.2`) enable nx to auto-resolve commit ranges for changelog and version calculation
- New packages can be added by setting the `npm:standalone` tag in their `project.json` and adding them to the `standalone` group in `nx.json`

### DIFF
--- a/.github/workflows/publish-standalone.yml
+++ b/.github/workflows/publish-standalone.yml
@@ -1,0 +1,159 @@
+name: Publish Standalone Packages
+
+on:
+  workflow_dispatch:
+    inputs:
+      dry-run:
+        description: 'Run in dry-run mode (no publishing, no git commits)'
+        required: false
+        default: false
+        type: boolean
+
+jobs:
+  publish-standalone:
+    # Prevent infinite loops: skip if the commit was made by the build system itself
+    if: "!contains(github.event.head_commit.author.email, 'flubuild@microsoft.com')"
+    name: Publish standalone packages
+    runs-on: ubuntu-latest
+
+    permissions:
+      contents: write
+      packages: write
+
+    steps:
+      - name: Validate branch
+        run: |
+          if [ "${{ github.ref }}" != "refs/heads/main" ]; then
+            echo "❌ ERROR: This workflow can only be run on the main branch" >> $GITHUB_STEP_SUMMARY
+            echo "" >> $GITHUB_STEP_SUMMARY
+            echo "Current ref: ${{ github.ref }}" >> $GITHUB_STEP_SUMMARY
+            exit 1
+          fi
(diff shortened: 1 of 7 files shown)


.github/workflows/publish-standalone.yml
```diff
@@ -0,0 +1,159 @@
...
+      dry-run:
+        description: 'Run in dry-run mode (no publishing, no git commits)'
+        required: false
+        default: false
+        type: boolean
+
+jobs:
+  publish-standalone:
+    # Prevent infinite loops: skip if the commit was made by the build system itself
+    if: "!contains(github.event.head_commit.author.email, 'flubuild@microsoft.com')"
```
mainframev, line 15: is `head_commit` available for workflow_dispatch?

mainframev (approved):

.github/workflows/publish-standalone.yml
Hotell: good question, not 100% sure, this is the same as we have in main publish. the thing is that this is not needed because we don't automatically run publish pipelines anymore. can be removed as follow-up

PR #1082, opened 2026-05-13
fix(react-icons-atomic-webpack-loader): skip indirect re-exports to avoid duplicate-export build failures

mainframev:
## Problem

The atomic webpack loader produces a duplicate export when a module re-exports a barrel-imported icon binding indirectly:

input:

```js
import { X } from '@fluentui/react-icons';
export { X };
```

output:

```js
export { X } from '@fluentui/react-icons/x';
export { X };
```

`oxc-parser` anchors the `StaticExport.{start,end}` of an indirect re-export to the originating `import` statement (not the `export` line), so the loader's `src.overwrite(exp.start, exp.end, …)` lands on the import range that the import-rewrite loop has already overwritten, while the original `export { X };` line is left untouched. Webpack then dies with:

```
Module parse failed: Duplicate export 'X'
```

This is reproducible against `@1js/fluent-icons-web/lib/types.js` (and other published packages — `@editor/settings`, `@office-iss/canvas-contextual-pal`) in OWA.

## Fix

Direct (`export { X } from 'mod';`) and indirect (`import { X } from 'mod'; export { X };`) re-exports are indistinguishable at the entry level — same `moduleRequest`, same `importName`/`exportName`/`localName` kinds. They differ only in where `oxc-parser` anchors the outer `StaticExport` record:

| Pattern | Outer record anchored at | Entry position |
|---|---|---|
| direct | the `export … from` statement | inside outer range |
| indirect | the originating `import` statement | outside outer range (points at the actual `export { X };` line) |

Add two clauses to the export-entry filter so we only rewrite entries whose own `[start, end]` falls inside their enclosing record's `[start, end]`:

```ts
e.start >= exp.start &&
e.end <= exp.end,
```

The import-rewrite loop already redirects the local binding to the atomic path, so the original `export { X };` continues to resolve correctly with zero extra code emitted.

## Test

Four fixtures exercise the indirect pattern and its variants; all fail pre-fix with `Module parse failed: Duplicate export …` and build cleanly post-fix:

- `test/src/indirect-reexports.js` — `export { AddFilled, ArrowLeftRegular, iconFilledClassName };`
- `test/src/aliased-indirect-reexports.js` — aliased import + plain export (`import { AddFilled as MyAdd … }; export { MyAdd, … };`)
- `test/src/renamed-indirect-reexports.js` — plain import + renamed export (`export { AddFilled as MyAdd, … };`)
- `test/src/mixed-reexports.js` — both indirect and direct re-exports in the same module

The `verify-transforms` plugin asserts the expected atomic paths (`@fluentui/react-icons/svg/add`, `…/svg/arrow-left`, `…/utils`) appear in each bundle and the barrel name (`"@fluentui/react-icons"`) does not.

### DIFF
--- a/packages/react-icons-atomic-webpack-loader/src/transform.ts
+++ b/packages/react-icons-atomic-webpack-loader/src/transform.ts
@@ -75,14 +75,11 @@ export function transformSource(source: string, options: TransformOptions): Tran
 
   for (const exp of staticExports) {
     const relevantEntries = exp.entries.filter(
-      (e) =>
-        e.moduleRequest?.value === MODULE_NAME &&
-        e.exportName.kind === 'Name' &&
-        e.start >= exp.start &&
-        e.end <= exp.end,
+      (e) => e.moduleRequest?.value === MODULE_NAME && e.exportName.kind === 'Name',
     );
-
     if (relevantEntries.length === 0) continue;
+
+    if (source.startsWith('import', exp.start)) continue;
 
     const lines: string[] = [];
 
--- a/packages/react-icons-atomic-webpack-loader/src/transform.ts
+++ b/packages/react-icons-atomic-webpack-loader/src/transform.ts
@@ -79,6 +79,9 @@ export function transformSource(source: string, options: TransformOptions): Tran
     );
     if (relevantEntries.length === 0) continue;
 
+    // Skip indirect re-exports (`import { X } from '...'; export { X };`): oxc reports these as static
+    // exports anchored at the originating `import` statement. The import loop above has already
+    // rewritten that range, so emitting again here would produce duplicate declarations.
     if (source.startsWith('import', exp.start)) continue;
 
     const lines: string[] = [];
--- a/packages/react-icons-atomic-webpack-loader/src/transform.ts
+++ b/packages/react-icons-atomic-webpack-loader/src/transform.ts
@@ -75,8 +75,13 @@ export function transformSource(source: string, options: TransformOptions): Tran
 
   for (const exp of staticExports) {
     const relevantEntries = exp.entries.filter(
-      (e) => e.moduleRequest?.value === MODULE_NAME && e.exportName.kind === 'Name',
+      (e) =>
+        e.moduleRequest?.value === MODULE_NAME &&
+        e.exportName.kind === 'Name' &&
+        e.start >= exp.start &&
+        e.end <= exp.end,
     );
+
     if (relevantEntries.length === 0) continue;
 
     const lines: string[] = [];
--- a/packages/react-icons-atomic-webpack-loader/test/webpack.config.js
+++ b/packages/react-icons-atomic-webpack-loader/test/webpack.config.js
@@ -39,6 +39,28 @@ const entries = {
     ],
     mustExclude: ['"@fluentui/react-icons"'],
   },
+  'aliased-indirect-reexports': {
+    src: './src/aliased-indirect-reexports.js',
+    loaderOptions: {},
+    mustInclude: ['@fluentui/react-icons/svg/add', '@fluentui/react-icons/svg/arrow-left'],
+    mustExclude: ['"@fluentui/react-icons"'],
+  },
+  'renamed-indirect-reexports': {
+    src: './src/renamed-indirect-reexports.js',
+    loaderOptions: {},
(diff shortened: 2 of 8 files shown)


packages/react-icons-atomic-webpack-loader/test/webpack.config.js
```diff
@@ -29,6 +29,38 @@ const entries = {
     ],
     mustExclude: ['"@fluentui/react-icons"'],
   },
+  'indirect-reexports': {
+    src: './src/indirect-reexports.js',
```
Hotell, line 33: it seems these kind of tests are not sufficient as this wont catch the problem this pr tries to fix right ? 

on the other hand I can't think of better approach as the output is already processed by webpack by the time the assertion runs which again creates different transform in comparison with the oxc one.

I guess we could directly test the output of our `transform` while keeping this suite as e2e coverage ?

packages/react-icons-atomic-webpack-loader/src/transform.ts
```diff
@@ -75,8 +75,13 @@ export function transformSource(source: string, options: TransformOptions): Tran
 
   for (const exp of staticExports) {
     const relevantEntries = exp.entries.filter(
-      (e) => e.moduleRequest?.value === MODULE_NAME && e.exportName.kind === 'Name',
+      (e) =>
+        e.moduleRequest?.value === MODULE_NAME &&
+        e.exportName.kind === 'Name' &&
+        e.start >= exp.start &&
```
mainframev: Skip indirect re-exports: oxc anchors them at the originating import, so entry positions fall outside the export record's own range

packages/react-icons-atomic-webpack-loader/src/transform.ts
```diff
@@ -75,8 +75,13 @@ export function transformSource(source: string, options: TransformOptions): Tran
 
   for (const exp of staticExports) {
     const relevantEntries = exp.entries.filter(
-      (e) => e.moduleRequest?.value === MODULE_NAME && e.exportName.kind === 'Name',
+      (e) =>
+        e.moduleRequest?.value === MODULE_NAME &&
```
Hotell: can't we just do  (bail if import was already rewritten ? )

```
// after  if (relevantEntries.length === 0) continue;
if (source.startsWith('import', exp.start)) continue;
``` 
?

packages/react-icons-atomic-webpack-loader/test/webpack.config.js
mainframev: hm, the webpack tests are failing without the fix: 

```bash
  ERROR in ./src/indirect-reexports.js 5:9
  Module parse failed: Duplicate export 'AddFilled' (5:9)

  ERROR in ./src/aliased-indirect-reexports.js 4:9
  Module parse failed: Duplicate export 'MyAdd' (4:9)

  ERROR in ./src/renamed-indirect-reexports.js 4:22
  Module parse failed: Duplicate export 'MyAdd' (4:22)

  ERROR in ./src/mixed-reexports.js 4:9
  Module parse failed: Duplicate export 'AddFilled' (4:9)
```

maybe it's sufficient to have current tests

packages/react-icons-atomic-webpack-loader/src/transform.ts
mainframev: Good idea, ty!

packages/react-icons-atomic-webpack-loader/test/webpack.config.js
Hotell: yes they fail on the webpack duplicate export which validates it compiles, but we have no actual visibility if the actual output is as expected beyond "relying" on if webpack compiled it it is correct right. we should consider the separate transformer test as additional safe guard

packages/react-icons-atomic-webpack-loader/src/transform.ts
```diff
@@ -79,6 +79,8 @@ export function transformSource(source: string, options: TransformOptions): Tran
     );
     if (relevantEntries.length === 0) continue;
 
+    if (source.startsWith('import', exp.start)) continue;
```
Hotell, line 85: maybe leave some comment what  is this for in the code ?

Hotell (approved):
left some replies, otherwise LGTM

PR #1095, opened 2026-06-03
feat(react-icons-file-type): implement v8 alternative for v9,modern

Hotell:
### Summary

Introduces a new package, `@fluentui/react-icons-file-type`, as a v9-compatible, dependency-light successor to the v8 `@fluentui/react-file-type-icons` package. It renders Fluent file type icons (document/spreadsheet/folder glyphs, etc.) as plain `` elements sourced from CDN-hosted assets — **without** a global icon registry and **without** hardcoding the CDN into application code.

### Motivation

The v8 `@fluentui/react-file-type-icons` package is built on v8 styling and v8 packages that are not compliant with v9, so it can't be used in v9 apps. This package is a v9-native replacement. The Fluent CDN is the default asset host in both v8 and v9; the difference here is that it's an overridable default of `FileTypeIconsProvider` rather than something you wire through your app. This package:

- Decouples from v8 (`<Icon>`, `registerIcons`, global icon registry) and v8 styling.
- Keeps the Fluent CDN as the default asset host while making it overridable via context — no need to configure a CDN in userland to get started.

### Public API

- **`<FileTypeIcon>`** — Griffel-styled `` component. Renders by file `extension` or by `type` (`FileIconType`) for icons not tied to an extension (folders, lists, etc.).
- **`<FileTypeIconsProvider>` / `useFileTypeIconsContext`** — configure the asset host (`baseUrl`) once for a subtree. `baseUrl` resolve precedence: `FileTypeIconsProvider` → Fluent CDN default.
- **`FileIconType`** — `const` object (not a TS `enum`) usable as both value (`FileIconType.folder`) and type, plus the `FileIconTypeInput` alias.
- **Constants** — `DEFAULT_BASE_URL`, `FLUENT_CDN_BASE_URL`, `DEFAULT_ICON_SIZE`, `ICON_SIZES`, and the `FileTypeIconSize` / `ImageFileType` types.
- **Headless entry point** — `@fluentui/react-icons-file-type/headless` ships the same `FileTypeIcon` with **no Griffel runtime**; consumers bring their own styles via the opt-in `headless/styles.css` or by targeting the `data-fui-filetype-icon` attribute. It re-exports the provider, `FileIconType`, and constants so it is usable standalone, and additionally exposes the composable `useFileTypeIcon` (state) / `renderFileTypeIcon` (render) hooks plus the `FileTypeIconState` type.

The extension→icon resolution (`getFileTypeIconSrc` / `getFileTypeIconSrcSet`) is an internal implementation detail, not a public export.

### SSR-safe high-DPI selection via `srcset`

Earlier iterations resolved the asset density in JS by reading `window.devicePixelRatio` (and threaded a `targetWindow` through the provider). That is **not SSR-safe**: the server has no `window`, so it always emits the 1x URL, and the first client render must match it — then a post-hydration effect swaps in the high-density asset, causing a flicker and a redundant download.

`FileTypeIcon` now resolves density in the **browser** via [`srcset`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/srcset): it renders a deterministic 1x `src` plus a `srcset` listing every published density and lets the browser pick. This is fully server/client deterministic (no hydration mismatch, no flicker, no double-fetch) and the browser re-evaluates automatically when the image moves to a display with a different DPI. As a result, `targetWindow` and the JS DPR resolver were removed.

```html
<!-- v9 output (png, size 24) -->
<img
  src="…/24/docx.png"
  srcset="…/24/docx.png 1x, …/24_1.5x/docx.png 1.5x, …/24_2x/docx.png 2x, …/24_3x/docx.png 3x, …/24_4x/docx.png 4x"
  width="24"
  height="24"
/>
```

### Implementation notes

> **NOTE:** bundle was reduced by ~2.5kB by packing the icon map and removing TS enums + verbose switch statements

- **Fluent v9 state/style/render architecture.** `FileTypeIcon` is composed from a shared `useFileTypeIcon` state hook (resolves props → `` slot props, including the `srcset`) and a `renderFileTypeIcon` render hook. The default entry point inserts a `useFileTypeIconStyles` style hook in between to layer the static Griffel styles; the `/headless` entry point composes the same state/render hooks with **no** Griffel runtime. Both render a single `` (no wrapper element).
- **Single ``, image-appropriate styling.** Styling targets the `` directly: `display: inline-block` and the replaced-element-only `object-fit: contain` (keeps non-square assets from stretching in the square `width`/`height` box). The v8 container-oriented `overflow: hidden` is dropped — it is a no-op on a replaced element.
- **`src/common/` layout.** Framework-agnostic/shared modules (resolver, constants, `FileIconType`, generated maps + JSON sources, context/provider, and the `useFileTypeIcon` hooks) live under `src/common/`, leaving only component wiring at the `src/` root and under `src/headless/`.
- **Codegen from declarative JSON.** Both the extension→icon table (`fileTypeIconMap.json`) and the `FileIconType` enum + its reverse lookup (`fileIconTypes.json`) are authored as declarative JSON sources of truth and compiled by build-time generators. The extension map ships as a minimal packed string (`icon:ext,ext;…`) with a lazy, cached decoder; `FileIconType` is generated as a `const` object (each entry declaring an explicit, validated numeric value) alongside a positional reverse-lookup array. Kept in sync via a `pretest`/`prebuild` `generate` step.
- **Bundle-size hygiene.** `FileIconType` is a `const` object (no enum reverse-map), the type→icon resolution uses a positional array instead of a 24-case `switch`, and the icon-src resolver composes the URL directly (no string encode/decode protocol) — trimming dead runtime weight.
- Dual ESM (`lib/`) + CJS (`lib-cjs/`) build.
- Vitest test suite (component a11y, headless component, provider, internal resolvers, map/enum decode, generated-file sync, and **SSR/hydration** tests asserting density-independent, mismatch-free markup on a high-DPI client).
- `bundle-size` (monosize) target with fixtures, including a fixed Griffel CSS-extraction config (adds the prerequisite `@griffel/webpack-loader` AOT transform so measurements reflect extracted CSS).
- Registered in the nx `standalone` release group.

### Docs

Adds a Storybook "File Type Icons" section to the docsite with **Overview, Sizes, Types, and Provider** stories, an **Overview** description (including the headless API), and a dedicated **Migration (v8 → v9)** page. The migration guide covers the SSR-safe `srcset` density handling and the single-`` styling change (no wrapper, `object-fit` on the image, no `overflow`).

### Migration highlights (v8 → v9)

| v8 | v9 |
| --- | --- |
| `initializeFileTypeIcons()` at startup | Not needed |
| `<Icon {...getFileTypeIconProps(options)} />` | `<FileTypeIcon {...options} />` |
| Custom CDN via `initializeFileTypeIcons(baseUrl)` | `<FileTypeIconsProvider baseUrl="...">` (Fluent CDN is the default) |
| Global icon registry | No global state |
| Single density-resolved `src` (not SSR-safe) | SSR-safe 1x `src` + `srcset`; browser picks the density |
| `<Icon>` wrapper with container styles (`overflow`) | Single ``, no wrapper; `object-fit` on the image |

### Testing

- `nx run react-icons-file-type:build` — emits `lib/` + `lib-cjs/`
- `nx run react-icons-file-type:lint` — passes
- `nx run react-icons-file-type:test` — all tests pass
- `nx run react-icons-file-type:bundle-size` — measures fixtures

### Bundle size

Measured via the `bundle-size` (monosize) target. The default `FileTypeIcon` is Griffel-styled; Griffel CSS extraction moves the static styles out of JS into a separate `griffel.css` asset (the JS number below is what monosize reports), so consumers that opt into Griffel build-time extraction get the smaller figure. The **headless** `FileTypeIcon` ships no Griffel runtime, so its size is the same with or without extraction.

**Default (no extraction)**

| Fixture | Minified size | GZIP size |
| --- | --- | --- |
| FileTypeIcon | 33.122 kB | 12.52 kB |
| FileTypeIcon (headless, no Griffel) | 4.664 kB | 2.468 kB |
| FileTypeIconsProvider | 286 B | 235 B |

**With Griffel extraction** (`GRIFFEL_EXTRACTION=true`)

| Fixture | Minified size | GZIP size |
| --- | --- | --- |
| FileTypeIcon | 7.992 kB | 3.994 kB |
| FileTypeIcon (headless, no Griffel) | 4.664 kB | 2.468 kB |
| FileTypeIconsProvider | 286 B | 235 B |


packages/react-icons-file-type/src/FileIconType.ts
```diff
@@ -0,0 +1,59 @@
+/**
+ * Enumerates special file type icons that do not map to any file extensions.
+ * For example, the 'pptx' icon maps to the extensions 'ppt', 'pptm', 'pptx',
+ * but the 'folder' icon does not map to any extensions and should be obtained
+ * via this enum.
+ */
+
+export enum FileIconType {
```
Hotell: this is from v8, ideally we should not use enums

packages/react-icons-file-type/src/FileTypeIconMap.ts
```diff
@@ -0,0 +1,594 @@
+/**
+ * Enumeration of icon file names, and what extensions they map to.
+ * Please keep items alphabetical. Items without extensions may require specific logic in the code to map.
+ * Always use getFileTypeIconProps to get the most up-to-date icon at the right pixel density.
+ */
+export const FileTypeIconMap: { [key: string]: { extensions?: string[] } } = {
```
Hotell: this is not tree-shakeable / needs improvement

packages/react-icons-file-type/src/FileTypeIcon.tsx
```diff
@@ -0,0 +1,68 @@
...
+  /**
+   * The image format to use.
+   * @default 'svg'
+   */
+  imageFileType?: ImageFileType;
+  /**
+   * Override the base URL used to resolve this icon's asset. When omitted, the `baseUrl`
+   * from the nearest `FileTypeIconsProvider` (or the Fluent CDN default) is used.
+   */
+  baseUrl?: string;
```
layershifter: Do we want to keep this? Provider should be config source

packages/react-icons-file-type/src/index.ts
```diff
@@ -0,0 +1,24 @@
...
+export {
+  getFileTypeIconProps,
+  getFileTypeIconNameFromExtensionOrType,
+  getFileTypeIconSuffix,
+} from './getFileTypeIconProps';
+export type { FileTypeIconOptions } from './getFileTypeIconProps';
+
+export { getFileTypeIconSrc } from './getFileTypeIconSrc';
+export { getFileTypeIconAsUrl } from './getFileTypeIconAsUrl';
+export { getFileTypeIconAsHTMLString } from './getFileTypeIconAsHTMLString';
```
layershifter: Is it needed for React impl?

packages/react-icons-file-type/README.md
```diff
@@ -0,0 +1,79 @@
...
+  getFileTypeIconAsHTMLString,
+  getFileTypeIconProps,
+} from '@fluentui/react-icons-file-type';
+
+getFileTypeIconAsUrl({ extension: 'docx', size: 24 }); // -> "https://.../24/docx.svg"
+getFileTypeIconAsHTMLString({ extension: 'docx', size: 24 }); // -> "<img ... />"
+getFileTypeIconProps({ extension: 'docx', size: 24 }); // -> { iconName, 'aria-label' }
+```
+
+All helpers accept an optional `baseUrl` second argument and default to the Fluent CDN.
```
layershifter: Image

Do we really want to have framework agnostic in a React package?

Image

packages/react-icons-file-type/src/getFileTypeIconProps.ts
```diff
@@ -0,0 +1,214 @@
...
+
+export function getFileTypeIconSuffix(
+  size: FileTypeIconSize,
+  imageFileType: ImageFileType = 'svg',
+  win?: Window,
+): string {
+  // SSR-safe: `window` may be undefined in non-DOM environments. Default the
+  // device pixel ratio to 1 (the standard-density asset) when it is unavailable.
+  // eslint-disable-next-line no-restricted-globals
+  const resolvedWindow = win ?? (typeof window !== 'undefined' ? window : undefined);
```
layershifter: We should use a context for `window`...

packages/react-icons-file-type/src/FileTypeIconMap.ts
```diff
@@ -0,0 +1,594 @@
...
+      'vsl',
+      'xla',
+      'xlam',
+      'xlb',
+      'xll',
+    ],
+  },
+  todoitem: {},
+  txt: {
+    extensions: ['dif', 'diff', 'readme', 'out', 'plist', 'properties', 'text', 'txt'],
```
layershifter: The structure is readable, but might not be good for bundle size

- First step to kill `extensions`: `{ extensions: [] }` => `[]`
- Then switch `{}` to `null`, `void`, `0`, drop empty keys?
- Flatten to array? I.e. `{ txt: { extensions: ['dif'] }` => `['txt', 'dif']` (`[key: string, ...extensions: string[]`)

WDYT?

packages/react-icons-file-type/src/index.ts
Hotell: most likely not

packages/react-icons-file-type/src/FileTypeIcon.tsx
Hotell: we can drop and re-add later if needed

packages/react-icons-file-type/src/FileTypeIcon.tsx
```diff
@@ -0,0 +1,68 @@
...
+    return null;
+  }
+
+  return (
+    <img
+      {...imgProps}
+      src={src}
+      width={size}
+      height={size}
+      alt={alt ?? extension ?? ''}
```
layershifter: Plz check that ARIA roles/attrs are aligned with `@fluentui/react-icons`.

packages/react-icons-file-type/README.md
Hotell: yeah I'll drop for now, the original intent was to keep v8 apis for v9

packages/react-icons-file-type/src/FileTypeIconMap.ts
Hotell: i went with packed format to reduce the bundle bloat to minimum

bigbadcapers:
hey @Hotell @layershifter thank you so much for putting this together, reviewing now! was trying to find alternative paths forwards to help get partners and us unstuck, as i'm about to complete requests to add 6 more filetype icons to the registry.

question for in terms of supporting apps who are still taking v8 drops and still a bit of runway before they can modernize: in terms of DRY principles, what does this mean for the icons-extension mappings i would be adding to this file: https://github.com/microsoft/fluentui/blob/master/packages/react-file-type-icons/src/FileTypeIconMap.ts do we now have to do that twice, or just a cut-over once?

Hotell:
> question for in terms of supporting apps who are still taking v8 drops and still a bit of runway before they can modernize: in terms of DRY principles, what does this mean for the icons-extension mappings i would be adding to this file: https://github.com/microsoft/fluentui/blob/master/packages/react-file-type-icons/src/FileTypeIconMap.ts do we now have to do that twice, or just a cut-over once?

yeah twice. we could expose that as public api in this pkg and put this package to v8 dependencies, but because the approach is different in this version we would need to adjust ( the whole object is packed to reduce bundle impact ).

the same applies for the cdn, although that might be difficult to reuse from this one as in v8 it's not in the icon package itself rather in style-utilities

layershifter (approved):

packages/react-icons-file-type/src/headless/styles.css
```diff
@@ -0,0 +1,15 @@
...
+ * `@fluentui/react-icons-file-type/headless`. Import this file once in your app to reproduce
+ * the default (Griffel) box behavior without the CSS-in-JS runtime:
+ *
+ *   import '@fluentui/react-icons-file-type/headless/styles.css';
+ *
+ * Or skip it and target `[data-fui-filetype-icon]` (or a `className`) with your own CSS.
+ */
+[data-fui-filetype-icon] {
+  display: inline-block;
+  overflow-x: hidden;
```
dmytrokirpa: does it actually work no img tag? it should probably be applied to a parent tag, but not for img.

packages/react-icons-file-type/src/FileTypeIconsContext.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * Provides the `baseUrl` used by descendant `FileTypeIcon` components to resolve
+ * their CDN-hosted assets. Omitting `baseUrl` falls back to the Fluent CDN, mirroring
+ * the zero-configuration ergonomics of the legacy `initializeFileTypeIcons` default
+ * without hardcoding the CDN inside the icon component itself.
+ */
+export const FileTypeIconsProvider: React.FC<FileTypeIconsProviderProps> = (props) => {
+  const { baseUrl = DEFAULT_BASE_URL, targetWindow = defaultTargetWindow, children } = props;
+  const value = React.useMemo<FileTypeIconsContextValue>(() => ({ baseUrl, targetWindow }), [baseUrl, targetWindow]);
```
dmytrokirpa: could we store the `devicePixelRatio` only in the context instead of the full window object?

packages/react-icons-file-type/src/fileTypeIconResolver.ts
```diff
@@ -0,0 +1,108 @@
...
+export function getFileTypeIconSrc(
+  options: FileTypeIconOptions,
+  baseUrl: string = DEFAULT_BASE_URL,
+  win?: Window,
+): string {
+  const { extension, size = DEFAULT_ICON_SIZE, type, imageFileType = 'svg' } = options;
+  const baseIconName = getFileTypeIconNameFromExtensionOrType(extension, type); // eg: docx
+  const densitySuffix = getFileTypeIconDensitySuffix(size, imageFileType, win); // eg: '' or '_3x'
+
+  return `${baseUrl}${size}${densitySuffix}/${baseIconName}.${imageFileType}`;
```
layershifter: Just realized, this repo doesn't have SSR tests. Is it going to return different results on a client and server?

Image

`getFileTypeIconDensitySuffix()` fallbacks to `1` when `window` is not defined

Hotell: good catch. not sure how to proceed, obviously the v8 version doesn't consider ssr at all AFAIR. suggestions ?

layershifter: Compute in provider and pass it down?

### Option A

Use always `1` as default, then catch up

Problem: state updates, flashes

```ts
function getDevicePixelRatio(targetWindow) {
  return Math.round((win?.devicePixelRatio ?? 1) * 10) / 10;
}

function Provider() {
  const [devicePixelRatio, setDevicePixelRatio] = React.useState(1);

  React.useEffect(() => { // should be useLayoutEffect, but it has per consequences due a sync update
    setDevicePixelRatio(getDevicePixelRatio(targetWindow));
   // ^ sync value on a client in a next update
  }, [])
}
```

### ~~Option 2~~

Prevent tree rendering till a value is there

Problem: part of a tree won't available from a server i.e. kills SSR

```
function Provider() {
  const [devicePixelRatio, setDevicePixelRatio] = React.useState(null);

  React.useEffect(() => {
    setDevicePixelRatio(getDevicePixelRatio(targetWindow));
   // ^ sync value on a client in a next update
  }, [])

  // to avoid flashes, we can do `devicePixelRatio: null | number`
  return devicePixelRatio === null ? null : <... />
}
```

### Option 3

https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/image/image-set

Problem: browser support

### Option 4

https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/srcset

No cons. Bigger HTML output, but

packages/react-icons-file-type/src/headless/FileTypeIcon.ssr.test.tsx
```diff
@@ -0,0 +1,117 @@
...
+const PNG_DENSITIES: ReadonlyArray<readonly [suffix: string, descriptor: string]> = [
+  ['', '1x'],
+  ['_1.5x', '1.5x'],
+  ['_2x', '2x'],
+  ['_3x', '3x'],
+  ['_4x', '4x'],
+];
+
+/** Runs `fn` with `window.devicePixelRatio` temporarily set to `value`, then restores it. */
+function withDevicePixelRatio<T>(value: number, fn: () => T): T {
```
layershifter: Is there sense in it when `srcset` is used?

packages/react-icons-file-type/src/FileTypeIconsContext.tsx
Hotell: pixel ration is handled by native browser primitives now

packages/react-icons-file-type/src/fileTypeIconResolver.ts
Hotell: option 4 implemented, ty!

packages/react-icons-file-type/src/headless/FileTypeIcon.ssr.test.tsx
Hotell: yeah , will remove

packages/docsite/stories/FileTypeIcons/FileTypeIconsSizes.stories.tsx
```diff
@@ -0,0 +1,32 @@
+import { makeStyles, shorthands } from '@fluentui/react-components';
+import { FileTypeIcon, ICON_SIZES } from '@fluentui/react-icons-file-type';
+import * as React from 'react';
+
+const useClasses = makeStyles({
+  container: {
+    display: 'flex',
+    alignItems: 'flex-end',
+    ...shorthands.gap('16px'),
```
dmytrokirpa, line 9:
```suggestion
    gap: '16px',
```

layershifter (approved):

packages/docsite/stories/FileTypeIcons/FileTypeIconsTypes.stories.tsx
```diff
@@ -0,0 +1,34 @@
+import { makeStyles, shorthands } from '@fluentui/react-components';
+import { FileTypeIcon, FileIconType } from '@fluentui/react-icons-file-type';
+import * as React from 'react';
+
+const useClasses = makeStyles({
+  container: {
+    display: 'flex',
+    alignItems: 'center',
+    ...shorthands.gap('16px'),
```
dmytrokirpa, line 9: same as above

dmytrokirpa (approved):

packages/docsite/stories/FileTypeIcons/FileTypeIconsTypes.stories.tsx
Hotell: will do as followup ty

PR #1096, opened 2026-06-08
feat(atomic-loader): support @fluentui/react-brand-icons

micahgodbolt:
## Problem

`@fluentui/react-brand-icons` (published from `microsoftdesign/Design System/_git/fluent-product-icons`) already ships atomic subpath exports (`./svg/*`), but the `@fluentui/react-icons-atomic-webpack-loader` is hardcoded to only transform `@fluentui/react-icons` barrel imports.

This forces consumers to either:
- Use barrel imports and pull the entire 340-icon bundle
- Manually rewrite every import to atomic paths
- Avoid the package entirely (which is what `m365-copilot-nav` in 1JS does today)

## Key Insight: The Packages Are Nearly Identical

| Feature | `react-icons` | `react-brand-icons` |
|---------|--------------|---------------------|
| Icon suffixes | Regular, Filled, Light, Color | Regular, Filled, Color |
| Atom path pattern | `./svg/<kebab-name>` | `./svg/<kebab-name>` |
| Utility exports | wrapIcon, bundleIcon, createFluentIcon, useIconState | Same |
| `sideEffects` | false | false |
| Suffix regex match | ✅ | ✅ (subset — same regex works) |

The existing `ICON_SUFFIX_REGEX` (`/(\d*)?(Regular|Filled|Light|Color)$/`) already matches brand icon names with zero changes.

## Proposed Solution

Extend the existing loader (not a new package) with a `modules` option:

### `src/transform.ts`
Replace hardcoded `MODULE_NAME = '@fluentui/react-icons'` with a configurable `Set` of module names. Pass the matched module name through to `getAtomicImportPath` so paths resolve correctly per-package.

### `src/index.ts`
Add `modules` option (default: both packages). Update fast-path string check to test against all configured modules.

### API (backward compatible)
```js
{
  loader: '@fluentui/react-icons-atomic-webpack-loader',
  options: {
    // Default: ['@fluentui/react-icons', '@fluentui/react-brand-icons']
    // Opt out of brand-icons:
    modules: ['@fluentui/react-icons'],
  },
}
```

### Before / After
```js
// Input
import { AddFilled } from '@fluentui/react-icons';
import { WordColor } from '@fluentui/react-brand-icons';

// Output (after loader)
import { AddFilled } from '@fluentui/react-icons/svg/add';
import { WordColor } from '@fluentui/react-brand-icons/svg/word';
```

## Implementation

I have a working implementation with all 25 tests passing (17 existing + 8 new). The diff is ~130 lines across 4 files:

- `src/transform.ts` — parameterize module name
- `src/index.ts` — add `modules` option, update fast-path
- `test/transform.test.ts` — 8 new tests
- `README.md` — document brand-icons + modules option

Happy to submit a PR if someone can point me to the right process (my EMU account can't fork this repo). I can share a `git format-patch` if helpful.

## Motivation

1JS consumers (M365 Copilot, etc.) need brand icons but can't adopt them without atomic import support — the bundle cost of the barrel import is prohibitive.

### DIFF
--- a/packages/react-icons-atomic-webpack-loader/src/transform.ts
+++ b/packages/react-icons-atomic-webpack-loader/src/transform.ts
@@ -2,29 +2,31 @@ import { parseSync } from 'oxc-parser';
 import type { StaticImport, StaticExport } from 'oxc-parser';
 import MagicString from 'magic-string';
 
-const MODULE_NAME = '@fluentui/react-icons';
+const DEFAULT_MODULES = ['@fluentui/react-icons', '@fluentui/react-brand-icons'];
 const ICON_SUFFIX_REGEX = /(\d*)?(Regular|Filled|Light|Color)$/;
 
 interface TransformOptions {
   iconVariant: 'svg' | 'fonts' | 'svg-sprite';
   path: string;
+  /** Icon package names to transform. Defaults to react-icons and react-brand-icons. */
+  modules?: string[];
 }
 
-function getAtomicImportPath(importName: string, iconVariant: 'svg' | 'fonts' | 'svg-sprite'): string {
+function getAtomicImportPath(importName: string, iconVariant: 'svg' | 'fonts' | 'svg-sprite', moduleName: string): string {
   if (importName === 'useIconContext' || importName === 'IconDirectionContextProvider') {
-    return '@fluentui/react-icons/providers';
+    return `${moduleName}/providers`;
   }
 
   const isIcon = importName.match(ICON_SUFFIX_REGEX);
 
   if (!isIcon) {
-    return '@fluentui/react-icons/utils';
+    return `${moduleName}/utils`;
   }
 
   const withoutSuffix = importName.replace(ICON_SUFFIX_REGEX, '');
   const kebabCase = withoutSuffix.replace(/[a-z\d](?=[A-Z])|[a-zA-Z](?=\d)|[A-Z](?=[A-Z][a-z])/g, '$&-').toLowerCase();
 
-  return `@fluentui/react-icons/${iconVariant}/${kebabCase}`;
+  return `${moduleName}/${iconVariant}/${kebabCase}`;
 }
 
 export interface TransformResult {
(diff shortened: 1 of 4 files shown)


packages/react-icons-atomic-webpack-loader/src/transform.ts
```diff
@@ -2,29 +2,31 @@ import { parseSync } from 'oxc-parser';
...
   path: string;
+  /** Icon package names to transform. Defaults to react-icons and react-brand-icons. */
+  modules?: string[];
 }
 
-function getAtomicImportPath(importName: string, iconVariant: 'svg' | 'fonts' | 'svg-sprite'): string {
+function getAtomicImportPath(importName: string, iconVariant: 'svg' | 'fonts' | 'svg-sprite', moduleName: string): string {
   if (importName === 'useIconContext' || importName === 'IconDirectionContextProvider') {
-    return '@fluentui/react-icons/providers';
+    return `${moduleName}/providers`;
```
Hotell, line 17: this is wrong, product icons has no provider, no utils export map

Image

packages/react-icons-atomic-webpack-loader/src/transform.ts
```diff
@@ -2,29 +2,31 @@ import { parseSync } from 'oxc-parser';
...
 
-const MODULE_NAME = '@fluentui/react-icons';
+const DEFAULT_MODULES = ['@fluentui/react-icons', '@fluentui/react-brand-icons'];
 const ICON_SUFFIX_REGEX = /(\d*)?(Regular|Filled|Light|Color)$/;
 
 interface TransformOptions {
   iconVariant: 'svg' | 'fonts' | 'svg-sprite';
   path: string;
+  /** Icon package names to transform. Defaults to react-icons and react-brand-icons. */
+  modules?: string[];
```
Hotell, line 12: this is bad design, product icons don't and wont support all `iconVariants`, so this api needs either better more generic design or handle the logic properly internally

Hotell (commented):
needs different design

Hotell:
also FYI the chunk is tree shakeable, the atomizer purpose is primarily for 2 things:
- preventing dupes with chunks and lazy loading scenarios
- bundler speed processing / dev loop

Hotell:
this is replaced https://github.com/microsoft/fluentui-system-icons/pull/1113 feel free to close @micahgodbolt

PR #1099, opened 2026-06-09
feat(react-icons): add image (CSS mask / <img/>) icon API

Hotell:
> **Draft / Alpha** — exploratory implementation of a new image-based icon API. Feedback welcome before stabilization.

## Summary

Adds a new **zero-JS-render** icon API under `@fluentui/react-icons/image`. Instead of shipping SVG path data in the JS bundle, each icon component carries only a short hashed **URL** to its `.svg`. Monochrome icons are drawn with CSS `mask-image` (recolorable via `currentColor`), and the deprecated multicolor `*_color` icons render as a native ``.

## Motivation

- **Remove icon rendering from JavaScript** — components render a `<span>` (mask) or `` (color); no SVG geometry in the JS bundle.
- **Smallest JS footprint** — under `asset/resource` the imported `.svg` compiles to a ~30–50 byte URL string per icon.
- **CSS-driven coloring** — mask icons inherit `color` (`background-color: currentColor`); `primaryFill` overrides it.
- **Strong caching** — each icon is a content-hashed, immutably cacheable `.svg`, fetched lazily in parallel over HTTP/2.
- **No CSS-in-JS runtime** — styling is a single opt-in vanilla CSS file (`image.css`).

## What's included

**Library (`packages/react-icons/src/image/`)**
- `createImageIcon` — monochrome mask `<span>` factory (`--fui-img` custom property + `currentColor`).
- `createImageColorIcon` — multicolor `` factory (a11y via `alt`).
- `image.css` — opt-in vanilla CSS: `mask`/`-webkit-mask` fallback, `currentColor`, RTL flip, and a forced-colors `CanvasText` override so masks stay visible in High Contrast Mode.
- `index.ts` / `shared.ts` — `@fluentui/react-icons/image` entry, types, constants.

**Generation & build**
- `scripts/image.writer.js` + `convert.js --image` emit **atoms only** (no chunks): one `.tsx` module per icon group plus one standalone **unfilled** `.svg` mask source per variant (color icons keep fills for ``).
- `build.js` copies image assets and registers the `./image` export map + `sideEffects`.

**Tests, docs, perf**
- `src/image/image.test.tsx` — render output, `--fui-img`, `currentColor`/`primaryFill`, no-SVG-in-DOM, sizing, RTL, color `` + `alt`, and the HCM CSS rule.
- `docs/preview-features/image.md` (+ README entry), marked alpha.
- `bundle-size/single-image.fixture.js` (+ `single-svg-sprite.fixture.js`).

**Showcase**
- `packages/icon-app` Image demo (recolor, sizing, color ``, a11y) + webpack `.css`/`.svg` loaders.

## Bundle size (single icon)

| API        | Minified | GZIP   |
| ---------- | -------- | ------ |
| **Image**  | 1.572 kB | 924 B  |
| SVG Sprite | 4.646 kB | 2.331 kB |
| Fonts      | 4.385 kB | 2.126 kB |
| SVG (JSX)  | 5.116 kB | 2.549 kB |

## Cross-browser / accessibility

- `mask-image` is Baseline 2023; `-webkit-mask` is shipped as the Safari < 15.4 fallback.
- `mask-image` cannot reliably reference SVG `#fragment` ids cross-browser, so each icon is served as its own whole `.svg` (not a `<use>` sprite).
- High Contrast Mode: masks pinned to `CanvasText` (would otherwise be forced to `Canvas` and become invisible).
- Color `` a11y mirrors the existing APIs via native `alt` (`title` → `alt`, else `alt=""`).

## Notes / scope

- **Out of scope (future):** a build-time subsetting plugin that inlines used icons into one CSS file (1 request) / `<style>` (0 requests) — only beneficial for small curated sets; per-icon hashed files are the default.
- The committed `package.json` currently also carries headless/sprite build-mutation entries that ride along with the generation enablement; these will be tidied before this leaves draft.
- Generated `lib/` and `src/atoms/image/` outputs are gitignored and produced by the build.

### DIFF
--- a/packages/icon-app/package.json
+++ b/packages/icon-app/package.json
@@ -14,7 +14,9 @@
   "devDependencies": {
     "@types/react": "19.1.13",
     "@types/react-dom": "19.1.9",
+    "css-loader": "^7.1.4",
     "html-webpack-plugin": "^5.6.3",
+    "style-loader": "^3.3.4",
     "ts-loader": "9.5.4",
     "typescript": "5.0.4",
     "webpack": "^5.100.2",
--- a/packages/icon-app/src/app.tsx
+++ b/packages/icon-app/src/app.tsx
@@ -4,6 +4,7 @@ import { makeStyles } from '@griffel/react';
 const Atoms = React.lazy(() => import('./atoms').then((mod) => ({ default: mod.Atoms })));
 const Chunk = React.lazy(() => import('./chunk').then((mod) => ({ default: mod.Chunk })));
 const All = React.lazy(() => import('./all').then((mod) => ({ default: mod.All })));
+const Image = React.lazy(() => import('./image').then((mod) => ({ default: mod.Image })));
 
 const useRootStyles = makeStyles({
   root: {
@@ -15,7 +16,7 @@ const useRootStyles = makeStyles({
 
 export function App() {
   const styles = useRootStyles();
-  const [state, setState] = React.useState<'atoms' | 'all' | 'chunk' | null>(null);
+  const [state, setState] = React.useState<'atoms' | 'all' | 'chunk' | 'image' | null>(null);
 
   return (
     <main>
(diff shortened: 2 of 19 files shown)


Hotell:
css mask add render perf overhead in comparison with sprites or font glyphs. closing

PR #1128, opened 2026-06-26
Update ic_fluent_trophy_16_filled.svg

julisbustoss2-pixel:


Hotell:
im not sure this is can land in, monochrome icons need to stay mono chrome.

PR #1130, opened 2026-06-29
feat: enforce conventional commits for commits and PR titles

Hotell:
## Summary

Enforces the [Conventional Commits](https://www.conventionalcommits.org/) standard for **both** local commit messages and PR titles, reusing a single shared config. This complements the existing nx `release` setup (`conventionalCommits: true`) so changelog and version bumps stay reliable.

## Changes

- **`commitlint.config.js`** (new) — shared config extending `@commitlint/config-conventional`. Standard types enforced; scope is optional/free-form. Typed via `@commitlint/types` JSDoc annotation.
- **`.husky/commit-msg`** (new) — validates every local commit message via commitlint.
- **`.github/workflows/pr.yml`** — new `valid-pr-title` job validates the PR title using the same config (title passed via env var to avoid shell injection).
- **`CONTRIBUTING.md`** — documents the new local commit-msg enforcement.
- **`package.json` / lockfile** — add `@commitlint/cli` + `@commitlint/config-conventional` dev dependencies.

## How it works

- **Locally:** after `npm install`, Husky's `commit-msg` hook rejects non-conventional commit messages.
- **In CI:** the `valid-pr-title` job fails the PR if its title is not a valid Conventional Commit.

## Verification

- `echo "bad message" | npx commitlint` → exits 1 (rejected)
- `echo "feat(react-icons): add new calendar icon" | npx commitlint` → exits 0 (accepted)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,6 +13,8 @@
         "importer"
       ],
       "devDependencies": {
+        "@commitlint/cli": "^21.1.0",
+        "@commitlint/config-conventional": "^21.1.0",
         "@griffel/webpack-extraction-plugin": "0.5.14",
         "@griffel/webpack-loader": "2.2.23",
         "@nx/js": "22.6.5",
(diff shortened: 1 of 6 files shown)


package-lock.json
tido64, line 1: If you're interested in a slightly more lightweight dependency, may I suggest [@rnx-kit/commitlint-lite
](https://github.com/microsoft/rnx-kit/blob/main/packages/commitlint-lite/README.md).

Hotell: if it provides 1:1 functionality happy to accept PR , ty

tudorpopams (approved):

Issue #936, opened 2025-12-11
[Bug]: Hiding an icon causes duplicate icons on the same page to become invisible


grhewitt:
### Package version

2.0.314

### React version

18.3.1

### Environment

```shell
I wasn't able to get this information as the command failed.
```

### Current Behavior

When two instances of the same icon are present on the page and then the first becomes hidden for some reason, the second icon becomes invisible.

This seems to be due to the fact that the icons are using the same ids for their gradients, so the second icon is reading gradient information from the first icon's defs. When the first icon becomes hidden the second icon renders with invisible colors.

### Expected Behavior

The second icon should not be affected by visibility changes to the first.

### Reproduction

https://stackblitz.com/edit/6fkxxb34?file=src%2Fexample.tsx

### Steps to reproduce

Use the toggle to hide and show the first icon, the second becomes invisible. You can also replace the duplicated icon with a different unique icon to see that the issue is exclusive to duplicated icons.

### Are you reporting an Accessibility issue?

no

### Suggested severity

Medium - Has workaround

### Are you willing to submit a PR to fix?

yes

### Validations

- [x] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.
- [x] The provided reproduction is a minimal reproducible example of the bug.

PR #1131, opened 2026-06-30
feat(react-icons): add idPrefix prop to scope Color variant SVG ids

Hotell:
## Summary

Adds an opt-in **`idPrefix`** prop to `Color` icon variants that scopes their locally-defined SVG `id`s (gradients, clip paths, filters) and the matching `url(#…)` references per rendered instance. This fixes the global DOM id collision where rendering the same color icon multiple times — and hiding one with `display: none` — breaks the gradients of the others ([#936](https://github.com/microsoft/fluentui-system-icons/issues/936)).

```tsx
<CalendarColor idPrefix="cal-a-" />
<CalendarColor idPrefix="cal-b-" />
```

## API

- New optional `idPrefix?: string` on `FluentIconsProps`.
- Applies to `Color` variants only; mono-color icons ignore it entirely.
- Stripped from the rendered DOM (never emitted as an attribute).

## Implementation notes

- A single `precomputeColorChildren(pathsOrSvg, options, prefix?)` performs the scoping inline during the existing `SvgNode` recursion (no extra tree clone; long path `d` strings short-circuit the `url(#…)` rewrite).
- A factory-time `createColorChildrenResolver` decides per **component** (an icon's "color-ness" is fixed) how children are produced:
  - **mono-color icons** → shared no-op, **no hook and no allocation** per instance;
  - **color icons** → a memoized hook keyed on `idPrefix`, so re-renders reuse the element tree and only rebuild when the prefix changes.
- Net result: zero added cost for the common (mono-color / no-prefix) path, O(1) shared precompute, and prefixed color icons don't rebuild their SVG tree on every re-render.
- Wired through both the Griffel (`utils`) and headless factories.

## Docs

- README: `idPrefix` documented as the recommended fix in the "SVG Gradient ID Collision" section, plus a note on `React.memo` as a complementary user-land optimization.
- Docsite: new **Icons/Color Variants** Storybook page with a side-by-side "without vs with `idPrefix`" example; User Guidance updated.

## Tests

- Added unit tests covering id/reference scoping, per-instance uniqueness across two instances, default (no-prefix) behavior, and the mono-color no-op.

## Why this approach (performance discussion)

The final design came out of working through a real tension between **per-instance memory** and **re-render cost**. The iterations:

1. **`useMemo` in every icon's render** — memoizes the scoped tree, but adds a hook slot (cached value + deps array) to *every* icon instance, including the ~99% mono-color icons that never use the feature. Rejected: a per-instance memory tax for a color-only edge case, multiplied across icon-heavy pages (1k+ icons).
2. **Plain ternary in render** (no hook) — zero overhead for mono-color icons, but prefixed color icons rebuild their whole SVG element tree on **every** re-render (no memoization). Rejected: degrades exactly the case the feature targets (many repeated color icons).
3. **Scoping inside `renderSvgBody`** — centralized and DRY, but since `renderSvgBody` runs every render it has the same "rebuild every re-render" problem as (2) for prefixed icons.

The constraint: you can't get *both* "no per-instance hook for mono-color" and "memoized scoping for prefixed color icons" from a single render-path branch, because hooks can't be called conditionally per render.

**Resolution — branch at factory time, not render time.** An icon's "color-ness" is fixed when the component is created, so `createColorChildrenResolver` decides once per component which strategy to install:

- **mono-color icons** → a shared no-op resolver: no hook is ever called, no per-component or per-instance allocation (identical cost to before the feature existed);
- **color icons** → a memoized `useMemo` hook keyed on `idPrefix`: re-renders reuse the element tree and only rebuild when the prefix actually changes.

Because the branch is fixed per component, hook order is stable across that component's renders, so it's Rules-of-Hooks-safe (verified with `eslint-plugin-react-hooks`). This satisfies **both** constraints: only color icons pay for the memo hook, and prefixed color icons no longer rebuild on every re-render.

### `React.memo` (complementary, user-land)

`React.memo(Icon)` is a separate, complementary lever and the bigger win when applicable — it skips the icon's render entirely (styles + state hooks + tree) when props are referentially stable. It's documented in the README as the recommended optimization for color-icon-heavy, frequently-updating trees. The two stack:

- `React.memo` → skips the whole render when **all** props are unchanged;
- internal `useMemo(idPrefix)` → reuses the scoped tree when the icon **does** re-render (e.g. `primaryFill` changed) but `idPrefix` did not.

We deliberately did **not** wrap icons in `React.memo` by default — shallow-compare across thousands of cheap icon renders isn't an obvious net win, and whether to memo is a consumer policy decision.

## Bundle size impact

This feature adds a **fixed ~353 B minified (~150–184 B gzip)** to every SVG icon fixture, plus a **~11 B minified (~10 B gzip)** bump to the font fixtures.

| Package & Exports | Baseline (min / gzip) | PR (min / gzip) | Change (min / gzip) |
|---|---|---|---|
| Atomic Imports | 7.125 kB / 3.285 kB | 7.478 kB / 3.435 kB | 🔺 353 B / 150 B |
| Headless - Bundle Icon | 2.767 kB / 1.437 kB | 3.12 kB / 1.621 kB | 🔺 353 B / 184 B |
| Headless - Single Fonts | 1.11 kB / 629 B | 1.121 kB / 639 B | 🔺 11 B / 10 B |
| Headless - Single SVG | 2.022 kB / 1.149 kB | 2.375 kB / 1.332 kB | 🔺 353 B / 183 B |
| Single Icon via Import all | 7.127 kB / 3.287 kB | 7.48 kB / 3.436 kB | 🔺 353 B / 149 B |
| Single Fonts | 8.133 kB / 3.533 kB | 8.144 kB / 3.543 kB | 🔺 11 B / 10 B |
| Single Icons | 7.127 kB / 3.287 kB | 7.48 kB / 3.436 kB | 🔺 353 B / 149 B |

### Why it grows

The `idPrefix` scoping logic — `renderSvgNode`'s id/`url(#…)` rewriting plus the per-instance memoized `createColorChildrenResolver` — lives in the **shared** `createFluentIcon` factory. Dead code inside a function body can't be tree-shaken based on call arguments, so **any** icon that imports `createFluentIcon` pulls in this color code. Hence the flat ~353 B on every SVG fixture.

### Why non-color (mono) variants are affected too

Mono icons are built by the **same** `createFluentIcon` factory, which statically references the color-rendering/scoping helpers (`renderSvgBody`'s color branch, `precomputeColorChildren`, `createColorChildrenResolver`). Importing even a single mono icon therefore bundles the whole factory — including code a mono icon never executes. That's why `Single Icons`, `Atomic Imports`, `Headless - Single SVG`, etc. all show the same +353 B, whether or not the icon has a color variant.

### Why font fixtures move by only ~11 B

Font icons bundle **no** SVG color code. Their small bump comes solely from the shared `useBaseIconState` hook, which now strips the `idPrefix` prop (one extra destructured binding) so it never lands on the DOM element. `createFluentFontIcon` routes through the same hook, so it inherits just that one-line cost.

## Follow-up — factory split (bigger change)

A stacked follow-up (`perf/react-icons-detach-color-factory`) **eliminates** this regression by splitting the factory into a lean `createFluentMonoIcon` (paths only, zero color code) and a `createFluentColorIcon` (`SvgNode[]` + `idPrefix`). Generated icons import only the factory they need, so mono icons stop bundling color code entirely — bringing them **below** the pre-feature baseline (e.g. `Headless - Single SVG` −149 B, `Single Icons` −126 B gzip). `createFluentIcon` is kept as a deprecated, fully backwards-compatible delegator. That change touches the generator and regenerates all icons, so it is split out for separate review.

## Notes

> This is a **draft** for review of the API shape and approach.

### DIFF
--- a/packages/docsite/stories/Icons/ColorVariants/index.stories.tsx
+++ b/packages/docsite/stories/Icons/ColorVariants/index.stories.tsx
@@ -1,0 +1,84 @@
+import { CalendarColor } from '@fluentui/react-icons';
+import { Body1Stronger, makeStyles, shorthands, tokens } from '@fluentui/react-components';
+import { Controls, Description, Primary, Title } from '@storybook/addon-docs/blocks';
+import * as React from 'react';
+
+import descriptionMd from './ColorVariantsDescription.md';
+
+const useClasses = makeStyles({
+  root: {
+    display: 'flex',
+    ...shorthands.gap('32px'),
+  },
+  column: {
+    display: 'flex',
+    flexDirection: 'column',
+    ...shorthands.gap('8px'),
+    ...shorthands.padding('12px'),
+    ...shorthands.border('1px', 'solid', tokens.colorNeutralStroke2),
+    borderRadius: tokens.borderRadiusMedium,
+  },
+  icons: {
+    fontSize: '48px',
+  },
+  // Hiding one instance with display:none triggers the gradient ID collision
+  // for any sibling that shares the same (unscoped) IDs.
+  hidden: {
+    display: 'none',
+  },
+});
+
+export const ColorIdPrefix = () => {
+  const classes = useClasses();
+
+  return (
+    <div className={classes.root}>
+      <div className={classes.column}>
+        <Body1Stronger>❌ Without idPrefix</Body1Stronger>
+        <div className={classes.icons}>
+          <CalendarColor className={classes.hidden} aria-hidden />
+          <CalendarColor aria-label="Calendar without idPrefix" />
+        </div>
+      </div>
+
+      <div className={classes.column}>
+        <Body1Stronger>✅ With idPrefix</Body1Stronger>
+        <div className={classes.icons}>
+          <CalendarColor idPrefix="cal-a-" className={classes.hidden} aria-hidden />
+          <CalendarColor idPrefix="cal-b-" aria-label="Calendar with idPrefix" />
+        </div>
+      </div>
+    </div>
+  );
+};
+
+export default {
+  title: 'Icons/Color Variants',
+  component: CalendarColor,
+  parameters: {
+    docs: {
+      description: {
+        component: descriptionMd,
+      },
+      // Single-story page: render the example once (Primary) and omit the
+      // autodocs `Stories` block, which would otherwise repeat the same story.
+      page: () => (
+        <>
+          <Title />
+          <Description />
+          <Primary />
+          <Controls />
+        </>
+      ),
+    },
+  },
+  argTypes: {
+    idPrefix: {
+      control: 'text',
+      description:
+        'Scopes the locally-defined SVG ids (gradients, clip paths, filters) of a `Color` variant per instance to avoid global DOM id collisions. Ignored by mono-color icons.',
+      table: { defaultValue: '' },
+      type: { name: 'string' },
+    },
+  },
+};
--- a/packages/react-icons/src/headless/createFluentIcon.ts
+++ b/packages/react-icons/src/headless/createFluentIcon.ts
@@ -3,7 +3,7 @@ import * as React from 'react';
 import type { FluentIconsProps } from './shared';
 import { iconClassName, cx } from './shared';
 import { useIconState } from './useIconState';
-import { computeViewBox, precomputeColorChildren, renderSvgBody } from '../core/svg';
+import { computeViewBox, createColorChildrenResolver, renderSvgBody } from '../core/svg';
 import type { SvgNode } from '../core/svg';
 
 export type FluentIcon = React.FC<FluentIconsProps>;
@@ -42,11 +42,12 @@ export const createFluentIcon = (
   options?: CreateFluentIconOptions,
 ): FluentIcon => {
   const viewBoxWidth = computeViewBox(width);
-  // Pre-render color SVG nodes once in the factory so the recursion
-  // never runs during React renders.
-  const colorChildren = precomputeColorChildren(pathsOrSvg, options);
+  // Resolve color children once per component: mono-color icons pay nothing,
+  // color icons get per-instance memoized `idPrefix` scoping.
+  const useColorChildren = createColorChildrenResolver(pathsOrSvg, options);
   const Icon = React.forwardRef((props: FluentIconsProps, ref: React.Ref<HTMLElement>) => {
     const iconState = useIconState(props, { flipInRtl: options?.flipInRtl });
+    const colorChildren = useColorChildren(props.idPrefix);
     const state = {
       ...iconState,
       className: cx(iconClassName, iconState.className),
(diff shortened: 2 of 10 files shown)


packages/docsite/stories/Icons/ColorVariants/index.stories.tsx
```diff
@@ -0,0 +1,84 @@
...
+  title: 'Icons/Color Variants',
+  component: CalendarColor,
+  parameters: {
+    docs: {
+      description: {
+        component: descriptionMd,
+      },
+      // Single-story page: render the example once (Primary) and omit the
+      // autodocs `Stories` block, which would otherwise repeat the same story.
+      page: () => (
```
Hotell, line 65: wanted to prevent the duplication of double story example that is forced by our index pattern.

packages/react-icons/src/headless/createFluentIcon.ts
```diff
@@ -42,11 +42,12 @@ export const createFluentIcon = (
...
   const viewBoxWidth = computeViewBox(width);
-  // Pre-render color SVG nodes once in the factory so the recursion
-  // never runs during React renders.
-  const colorChildren = precomputeColorChildren(pathsOrSvg, options);
+  // Resolve color children once per component: mono-color icons pay nothing,
+  // color icons get per-instance memoized `idPrefix` scoping.
+  const useColorChildren = createColorChildrenResolver(pathsOrSvg, options);
   const Icon = React.forwardRef((props: FluentIconsProps, ref: React.Ref<HTMLElement>) => {
     const iconState = useIconState(props, { flipInRtl: options?.flipInRtl });
+    const colorChildren = useColorChildren(props.idPrefix);
```
Hotell, line 50: only color variant that specifies idPrefix pays the memo/perf tax

packages/docsite/stories/Icons/ColorVariants/index.stories.tsx
```diff
@@ -0,0 +1,84 @@
...
+        <div className={classes.icons}>
+          <CalendarColor className={classes.hidden} aria-hidden />
+          <CalendarColor aria-label="Calendar without idPrefix" />
+        </div>
+      </div>
+
+      <div className={classes.column}>
+        <Body1Stronger>✅ With idPrefix</Body1Stronger>
+        <div className={classes.icons}>
+          <CalendarColor idPrefix="cal-a-" className={classes.hidden} aria-hidden />
```
dmytrokirpa, line 47: q: should we recommend using useId?

dmytrokirpa (approved):

packages/docsite/stories/Icons/ColorVariants/index.stories.tsx
Hotell: added to tips https://github.com/microsoft/fluentui-system-icons/pull/1131/changes#diff-35d3e7fcc2cb3fa91585228ffb109d2cc7b4bc584b0d1dc12e3986a7216a7297R18

PR #1136, opened 2026-07-02
docs(docsite): fix ColorVariants story 'Show code' to follow index re-export pattern

Hotell:
## Problem

The **Show code** panel on the *Icons / Color Variants* story was displaying the **entire story module** — the `default` meta, `argTypes`, and the custom `docs.page` render — instead of just the example component.

## Root cause

The Fluent `@fluentui/react-storybook-addon-export-to-sandbox` addon shows the *whole story module source* (so it can be exported to a sandbox). The rest of the docsite avoids this by following an **index re-export pattern**: each story lives in its own dedicated `*.stories.tsx` file containing only that example, while `index.stories.tsx` re-exports the stories and holds the meta/`argTypes`/page boilerplate.

`ColorVariants/index.stories.tsx` broke this by defining the `ColorIdPrefix` story inline alongside all the meta config, so everything leaked into the code display.

## Fix

- Add `ColorVariants/ColorVariants.stories.tsx` containing only the `ColorIdPrefix` story (imports, styles, component). This is now the clean source shown in **Show code**.
- Rewrite `ColorVariants/index.stories.tsx` to re-export the story and hold the `default` meta/`argTypes`/`page` config, mirroring `Icons/index.stories.tsx` and `FileTypeIcons/index.stories.tsx`.

The sibling `ColorVariants.stories.tsx` is not double-registered because the Storybook glob only matches `**/index.stories.@(ts|tsx)`.

## Verification

- `npx nx run docsite:build` completes successfully (Storybook + SWC compile all stories with no errors).

### DIFF
--- a/packages/docsite/stories/Icons/ColorVariants/ColorVariants.stories.tsx
+++ b/packages/docsite/stories/Icons/ColorVariants/ColorVariants.stories.tsx
@@ -1,0 +1,50 @@
+import { CalendarColor } from '@fluentui/react-icons';
+import { Body1Stronger, makeStyles, shorthands, tokens } from '@fluentui/react-components';
+import * as React from 'react';
+
+const useClasses = makeStyles({
+  root: {
+    display: 'flex',
+    ...shorthands.gap('32px'),
+  },
+  column: {
+    display: 'flex',
+    flexDirection: 'column',
+    ...shorthands.gap('8px'),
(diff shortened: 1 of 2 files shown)


Hotell:
replaced by https://github.com/microsoft/fluentui-system-icons/pull/1138

PR #1145, opened 2026-07-08
fix(react-icons): reserve 1em box for font icons to prevent layout shift

Hotell:
## Summary

Font icons suffered **two** distinct layout shifts (CLS). This PR fixes both, and adds docsite coverage for the font-icon variant.

1. **Horizontal shift** — font icons had no explicit box, so their footprint was derived from the glyph's advance width, which is `0` until the webfont paints (the fonts use `font-display: block`, and the glyphs are PUA codepoints with no fallback). The `<i>` collapsed to ~0 width, then snapped to the glyph width on load.
2. **Inline vertical shift** — even once a box is reserved, a font icon is an `inline-block` whose baseline is derived from its glyph. In normal inline text flow that baseline moves when the webfont's metrics swap in on load, growing the line box and pushing surrounding text down. (Flex chips like the other stories hide this via `align-items: center`; it only surfaces in inline text — see #1145 review discussion.)

## Fix 1 — reserve a `1em` box (horizontal CLS)

- `src/utils/fonts/createFluentFontIcon.styles.ts` (Griffel) + `src/headless/styles.css` (headless) — add `width: 1em; height: 1em` to the font-icon root, so a square box is reserved from first paint (mirrors the resizable SVG).
- Verified against the generated TTFs (`unitsPerEm = 500`): the Resizable font advance is exactly `1.000em`, and glyph *ink* never exceeds ~1.6% of the em box in any font.

## Fix 2 — pin the inline baseline (vertical CLS)

- Add `overflow: hidden` to the same font-icon root. This moves the `inline-block` baseline to its **bottom margin edge** (font-independent, exactly like the resizable SVG replaced element), so the box no longer depends on the glyph's baseline and stops shifting when the webfont loads.
- **No visible reposition:** the bottom edge is where the loaded FluentSystemIcons glyph already sits, so already-rendered (loaded) layouts don't move — measured 0px reposition vs today. Only the transient pre-paint state is corrected.
- **Why `overflow: hidden` specifically:** it's the only mechanism that relocates the baseline. `overflow: clip` (+`overflow-clip-margin`) does **not** change the baseline (verified — still shifts), and `vertical-align` keywords fix it but reposition the glyph ~6px (a visible breaking change). `hidden` clips the ~1.6% ink noted above, but that's sub-pixel at real sizes (~0.4px @24px, ~0.8px @48px) at the extreme corners only.
- Applied to **both** the Griffel styles and the headless CSS; `build-verify` and factory snapshots updated.

## Docsite (`packages/docsite`)

New **Icons/Font Icons** section (ordered after Color Variants) with self-contained stories:

- **Font Icons** — gallery of Regular/Filled glyphs.
- **Sizing** — side-by-side table comparing the *resizable* variant scaled via `font-size` against the purpose-built *sized* variant per size.
- **Box** — labelled SVG-vs-font comparison showing the identical `1em` footprint.
- **Layout Shift** — headless font icons with a shared box + guide illustrating the horizontal CLS the `1em` box prevents, plus a recorded GIF.
- **Inline Layout Shift** *(new)* — before/after regression demo of Fix 2 in normal inline text flow: the shipped default (`overflow: hidden`) stays put while a column opted out via `overflow: visible` drops on load, plus a recorded GIF.

font-icons-cls-demo

Supporting build wiring:

- **Font subsetting** — wired `@fluentui/react-icons-font-subsetting-webpack-plugin` into Storybook (`optimization.usedExports`), dropping the shipped icon fonts from ~10 MB to ~12 KB.
- **CSS extraction** — a small `webpackFinal` helper extracts *only* the headless icon CSS to a static `<link>` in the production build (so the Layout Shift demos' `@font-face` loads as real CSS with real `font-display: block`); everything else keeps Storybook's default `style-loader`.
- Long descriptions extracted to sibling `.md` files; the Font Icons intro links to the **Build Transforms** page (atomic loader + font subsetting tooling).

## Testing

- `nx run react-icons:test` and `nx run react-icons:build-verify` pass (snapshots updated for the `overflow` atom).
- Docsite `lint` + `storybook build` pass; verified the extracted `.css` assets and subset font sizes in the build output.
- Both fixes verified live in Storybook via throttled reloads (horizontal + inline vertical).

&gt; Note: the `fontSize`-prop parity for font icons is a separate change tracked in #1146.


packages/react-icons/src/utils/fonts/createFluentFontIcon.styles.ts
```diff
@@ -58,6 +58,13 @@ export const useStaticStyles = makeStaticStyles(`
 export const useRootStyles = makeStyles({
   root: {
     display: 'inline-block',
+    // Reserve a 1em box up front (mirrors the resizable SVG's width/height="1em").
+    // Without this the box is sized by the glyph advance, which is 0 until the
+    // webfont loads (font-display: block + PUA codepoints have no fallback glyph),
+    // causing layout shift on every font icon. Glyph ink stays within ~1.6% of the
+    // em box across all fonts, so overflow is intentionally left visible (no clip).
+    width: '1em',
+    height: '1em',
```
mainframev, line 66: in storybook story the icon rendered inside flex container and there is no CLS, but there might be still CLS if icon is rendered in inline text. Perhaps we need to double check it (I asked the agent to create a similar repro): 

Image

Hotell: excellent catch, we didn't run into this in modern as we set/override icon slots to flex 

- seems only solution is to use overflow https://github.com/microsoft/fluentui-system-icons/pull/1145/changes/761b49dd79e44775496bc4daa9f0434b33f2f79b
- the other option we have to keep this as is as the CLS is minor (only the glyph box moves )

wdyt?

mainframev: yeah, I think that works, I initially planned to suggest vertical-align, but you already covered with explanation, ty

mainframev (approved):

PR #1146, opened 2026-07-08
fix(react-icons): honor fontSize prop on resizable font icons (SVG/font parity)

Hotell:
## Summary

The `fontSize` prop resized **SVG** icons (it's a valid SVG `font-size` presentation attribute, and the icon is `1em`) but was a **no-op on font icons** — the font factory only applied its baked-in size, and any `fontSize` prop leaked onto the `<i>` as an inert HTML attribute.

That breaks a real scenario: when an icon is switched from the SVG to the font variant (e.g. via a bundler loader / barrel import), a consumer's `fontSize` prop silently stops sizing it.

This PR makes the prop behave consistently across both variants.

## Behavior

Both font factories (Griffel and headless) now map a `fontSize` prop to the `font-size` CSS style. To keep the SVG↔font swap transparent, the rule mirrors the SVG side exactly:

| variant | `fontSize` prop | SVG | font |
|---|---|---|---|
| **resizable** (`SendRegular`) | honored | `1em` scales with `font-size` | `style.fontSize` |
| **sized** (`Send24Regular`) | **ignored** | hardcoded `width`/`height` px | baked-in design size |

Only **resizable** font icons honor the prop. **Sized** font icons keep their baked-in design size — matching sized SVG icons, whose hardcoded `width`/`height` already ignore `font-size`. So an (accidental) `fontSize` on a sized icon is a no-op in *both* variants, and the swap stays visually identical.

## Changes

- `src/utils/fonts/createFluentFontIcon.tsx` + `src/headless/fonts/createFluentFontIcon.tsx` — pull `fontSize` off props and apply it via `applyFontStyle`, gated on the icon being resizable (`bakedFontSize === undefined ? propFontSize : bakedFontSize`).
- `src/core/fontIcon.ts` — `applyFontStyle` accepts `number | string`.
- `src/utils/FluentIconsProps.types.ts` — documents `fontSize?: string | number` (with a note that sized variants ignore it).
- `src/utils/icon-factories.test.tsx` — tests: resizable applies the prop (and doesn't leak it as an attribute); sized ignores it.

## Testing

- `nx run react-icons:test` — 90 passing (icon-factories: 13).
- `nx run react-icons:build-verify` — 135 passing.

## Note

Split out from #1145 (the font-icon `1em` box / CLS fix); this change is independent of it.

### DIFF
--- a/packages/react-icons/src/utils/icon-factories.test.tsx
+++ b/packages/react-icons/src/utils/icon-factories.test.tsx
@@ -57,6 +57,34 @@ describe('React component tests', () => {
     `);
   });
 
+  test('createFontIcon applies the `fontSize` prop as a CSS style (API parity with SVG icons)', () => {
+    const AccessTimeRegular: FluentFontIcon = createFluentFontIcon('AccessTimeRegular', '', 2, undefined);
+
+    const { container } = render(<AccessTimeRegular fontSize="32px" />);
+    const i = container.querySelector('i');
+
+    expect(i).toBeTruthy();
+    expect(i).toHaveStyle({ fontSize: '32px' });
+    // The prop must NOT leak onto the element as an (inert) DOM attribute.
+    expect(i).not.toHaveAttribute('fontSize');
+    expect(i).not.toHaveAttribute('fontsize');
+  });
+
+  test('createFontIcon sized icons ignore the `fontSize` prop and keep their baked-in size (parity with sized SVG)', () => {
+    // A non-undefined baked size marks a *sized* icon (e.g. Send24Regular).
+    const Send24Regular: FluentFontIcon = createFluentFontIcon('Send24Regular', '', 1, 24);
+
+    // No prop -> baked-in size is used.
+    const withDefault = render(<Send24Regular />).container.querySelector('i');
+    expect(withDefault).toHaveStyle({ fontSize: '24px' });
+
+    // `fontSize` prop is ignored: sized icons are fixed-size, matching sized SVG icons
+    // (whose hardcoded width/height ignore font-size), so a SVG->font swap stays consistent.
+    const withProp = render(<Send24Regular fontSize={16} />).container.querySelector('i');
+    expect(withProp).toHaveStyle({ fontSize: '24px' });
+    expect(withProp).not.toHaveAttribute('fontsize');
+  });
+
   test('createFluentIcon renders an SVG with expected path', () => {
     const d = 'M1 2 L3 4';
     const MyIcon = createFluentIcon('MyIcon', '1em', [d]);
--- a/packages/react-icons/src/core/fontIcon.ts
+++ b/packages/react-icons/src/core/fontIcon.ts
@@ -10,7 +10,7 @@ import * as React from 'react';
 export const applyFontStyle = (
   state: { style?: React.CSSProperties },
   primaryFill: string | undefined,
-  fontSize: number | undefined,
+  fontSize: number | string | undefined,
 ): void => {
   if (primaryFill && primaryFill.toLowerCase() !== 'currentcolor') {
     state.style = {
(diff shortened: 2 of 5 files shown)


packages/react-icons/src/utils/icon-factories.test.tsx
```diff
@@ -57,6 +57,34 @@ describe('React component tests', () => {
...
+    // A non-undefined baked size marks a *sized* icon (e.g. Send24Regular).
+    const Send24Regular: FluentFontIcon = createFluentFontIcon('Send24Regular', '', 1, 24);
+
+    // No prop -> baked-in size is used.
+    const withDefault = render(<Send24Regular />).container.querySelector('i');
+    expect(withDefault).toHaveStyle({ fontSize: '24px' });
+
+    // `fontSize` prop is ignored: sized icons are fixed-size, matching sized SVG icons
+    // (whose hardcoded width/height ignore font-size), so a SVG->font swap stays consistent.
+    const withProp = render(<Send24Regular fontSize={16} />).container.querySelector('i');
```
dmytrokirpa, line 83: is `<Send24Regular style={{ fontSize: 16} } />` also supported?

Hotell: using `style` was always supported for both svg/font. 

this PR is merely about unification of `fontSize` prop behaviour 

**Before:**

```
// svg
<AccessTimeFilled fontSize="64px" /> -> will scale to 64px

// font
<AccessTimeFilled fontSize="64px" /> -> will do nothing 🚨 ( if used via deep import TS will error as `fontSize` is not available directly on props API )
```

**After:**

```
// svg
<AccessTimeFilled fontSize="64px" /> -> will scale to 64px

// font
<AccessTimeFilled fontSize="64px" /> -> will scale to 64px
```

**sized icons:**

both `fontSize` and `style={{fontSize:''}}` are noop for both svg/font variant


Image

Hotell: I understand it's confusing, ideally sized icons `Props` should not contain `fontSize` but that would be a breaking change for all existing svg icon users. 

we could log error in runtime or ship lint rule. thoughts ?

dmytrokirpa: I didn't check the previous implementation, but I'd expect all icons (including "sized") to be resizable. And the only difference between "sized" and "resizable" that there is that "sized" icons have predefined size, but still possible to override

dmytrokirpa: But if that will be a breaking change, lets keep it as it is for now

dmytrokirpa: Also could we use `size` prop instead of `fontSize`, eg `<AccessTimeFilled fontSize="64px" /> => <AccessTimeFilled size={64} />`

Hotell: > only difference between "sized" and "resizable" that there is that "sized" icons have predefined size

the svg path/viewbox and font glyph are different so making those resizable would break fidelity.  

icons v1 versions didnt have resizable icons, v2 added that and it should be a default api to use for everyone


**from existing context:**

```
@fluentui/react-icons has been major bumped to 2.x. There are some key changes in the API and usages. The main ones are:

- The general icons are now resizeable using styling (i.e. scaling using the `height` and `width` props or using `fontSize` prop, etc). Ex: `<AccessTimeFilled fontSize={40}>`
- There is now a more general icon offering included with the sized icons. Instead of importing `<AccessTime24Filled>`, you just import `<AccessTimeFilled>`.
- The sized icons are still available, and are recommended if you know the specific size you want your icons to be. For the general case, the more general icons are available to you.
```

**what I suggest:**

- getting user base data  and reasoning why they used sized variants
- lint rule 'preferred-resizable` that would enforce resizable icons usage

Hotell: > `<AccessTimeFilled size={64} />`

- this would be an alias for `fontSize` prop basically ?
- its out of scope for this PR change but happy to discuss/ship

dmytrokirpa: > its out of scope for this PR change but happy to discuss/ship

absolutely, lets discuss that offline

> this would be an alias for fontSize prop basically ?

I was thinking about replacing all `width/height/fontSize` with just `size`, but agree that it needs more thinking and discussions

dmytrokirpa (approved):

PR #1148, opened 2026-07-09
ci(prerelease): scope pre-release to selected packages via projects input

Hotell:
## Summary

Adds the ability to pre-release an explicitly selected set of **already-public** packages via the `Publish Prerelease NPM` workflow, instead of the previously fixed `react-*` set. Each selected package is versioned and published **independently**.

## What changed

- **Workflow** (`.github/workflows/publish-prerelease.yml`): new required `projects` input (Nx selection syntax — names, globs, `tag:` selectors) + optional `exclude`. Private packages (`tag:npm:private`) are excluded by default.
- **`calculate-prerelease-version.js`**: computes each selected package's next prerelease version **independently** from the npm registry (no single-shared-version collapse) and writes them to `tmp/prerelease-versions.json`.
- **`apply-prerelease-versions.js`**: writes the computed `version` into each selected package's `package.json` (never touches `devDependencies`); supports `--dry-run=true|false` and surfaces a version-plan table in the run summary.
- **Skill** (`pre-release-npm-package`): single public-only flow — removed the "make private packages publishable" mode and the discovery script.
- **Docs** (`packages/react-icons/docs/prerelease.md`): updated for projects-based selection.

## Design notes

- **Public-only:** only already-public packages are eligible (private excluded). An already-released package is by definition already in an `nx.json` release group, which is what `nx release publish` requires — so no release-group edits are needed.
- **No dependency cascade:** each package is pre-released independently. A pre-released package keeps referencing the already-published version of its repo dependencies; to have a dependent consume a new prerelease, select it explicitly too.
- **Why not `nx release version`:** it enforces `preserveMatchingDependencyRanges` and rejects a prerelease that falls outside the repo's `*` devDependency ranges. The custom apply step writes only `version` and leaves `devDependencies` untouched, and `nx release publish` does not validate those ranges (verified end-to-end).

## Validation

- Per-project independent versions confirmed (`@fluentui/react-icons@2.0.333-alpha.0` + `@fluentui/react-icons-file-type@0.0.2-alpha.0`).
- `--dry-run=true|false` both behave correctly; workflow YAML validated.
- Applied a prerelease version to a package carrying a `"@fluentui/react-icons": "*"` devDependency, then `nx release publish --dry-run` → **succeeded** (publish does not validate devDep ranges).

> Draft — validated locally; an end-to-end run is pending via `workflow_dispatch` on a non-`main` branch (the workflow blocks `main` by design).


.github/skills/pre-release-npm-package/SKILL.md
```diff
@@ -1,16 +1,18 @@
...
+- **Publishable packages only** — Pre-release existing public packages. No code changes needed; the skill validates the branch and guides the user to trigger the CI workflow, where the package(s) are selected via the workflow's `projects` input.
+- **Include private packages** — Make private library packages publishable (removing their `"private": true` flag) so they can be pre-released. Code changes are committed on the pre-release branch.
 
-After this skill completes, the `Publish Prerelease NPM` GitHub Actions workflow (`.github/workflows/publish-prerelease.yml`) can be triggered to build, version, and publish the packages to npm.
+> **Note:** The `Publish Prerelease NPM` workflow selects packages ad-hoc via its `projects` input (Nx selection syntax) and excludes `tag:npm:private` by default. A package therefore only needs to be **publishable** (no `"private": true`) to be pre-released — it does **not** need to be added to a specific Nx release release group.
```
dmytrokirpa: Could you help me confirm my understanding here? This note (and the follow-ups at lines 89 and 103) says a package "does **not** need to be added to a specific Nx release group" — but the publish step runs `npx nx release publish --projects="$NX_NAMES"`, and as far as I can tell `nx release publish` only operates on projects that belong to a group under `nx.json > release.groups`.

When I tried it locally against a project that isn't in any group, it bailed out:
```
$ npx nx release publish --projects=docsite --dry-run=true; echo EXIT=$?
NX  ... did not match any configured release groups: - docsite
EXIT=1
```
I think this passes today only because every non-app package already happens to be in a group (`react`, `native`, `standalone`). My worry is the forward-looking case the PR mentions (a new `react-core` consumed by `react-brand-icons`): the workflow would build, version, and rewrite the package's `package.json`, then fail at publish with exit 1 — and take the rest of the selected packages down with it.

Am I missing something? If not, maybe we could note that a newly-publishable package also needs to be added to a release group in `nx.json`, or add a fail-fast check that gives a clear error when a selected project isn't in a configured group. Curious what you think!

.github/skills/pre-release-npm-package/SKILL.md
```diff
@@ -78,87 +80,36 @@ If no eligible packages are found, inform the user and fall back to "Published p
...
+Ask the user which eligible packages to make publishable for the pre-release. They may select one or more.
 
-If any selected package's Nx project name (the directory name under `packages/`) does **NOT** start with `react-`, warn the user:
+#### Workflow Compatibility Note
 
-> ⚠️ Package `<package-name>` has Nx project name `<dir-name>` which does not match the `react-*` glob used by the publish-prerelease workflow. This package will be added to the release group but will NOT be picked up for building/publishing unless the workflow is also updated.
+The `publish-prerelease.yml` workflow selects packages via its `projects` input using Nx project-selection syntax and **excludes `tag:npm:private` by default**. Any package the user names in that input will be picked up — there is no `react-*` naming restriction — **as long as it is publishable** (no `"private": true`).
 
-Let the user decide whether to proceed despite the warning.
+Therefore, the only change required to pre-release a currently-private package is removing its `"private": true` flag (Step 4). No Nx release-group membership is required.
```
dmytrokirpa: Same question as on the note up top — I think "No Nx release-group membership is required" might not hold, since `nx release publish --projects` seems to require each selected project to be in a release group in `nx.json` (otherwise it exits 1). Could you double-check this one?

dmytrokirpa (approved):

.github/skills/pre-release-npm-package/SKILL.md
Hotell: this is definitely an issue which the skill addresses ( needs tweaking )

basically if package is `private` and needs to be released as initial pre-release we will need to add it to some ephemeral temporary release group in order to make this work.

or we can save ourselves the trouble and don't allow pre-release on non public packages -> which IMO is the right call in terms of flow simplicity taking into consideration that we already have ZERO based semver in place. 

**wdyt : ?**

- only already published package can be target for pre-release
  - this also mitigates the confusion, because ATM if we pre-release private npm will add it "latest" tag which is very confusing

dmytrokirpa: > only already published package can be target for pre-release

Yes, that works and clears things up

Hotell: implemented. ty!

.github/skills/pre-release-npm-package/SKILL.md
Hotell: closing as resolved via previous convo. ty

dmytrokirpa (approved):

PR #1156, opened 2026-07-10
feat(file-icons): add wikiPage icon type

bigbadcapers:
Adds the wikiPage file type mapping on top of the file type icon package work. Draft until the upstream CDN/package flow is approved and ready.


Hotell (approved):
needs conflict resolution, otherwise LGTM

PR #1164, opened 2026-07-14
chore(eslint-plugin-react-icons): scaffold ESLint plugin package

Hotell:
## Summary

Scaffolds a new ESLint plugin package `@fluentui/eslint-plugin-react-icons`, modeled after [`@fluentui/eslint-plugin-react-components`](https://github.com/microsoft/fluentui/tree/master/packages/react-components/eslint-plugin-react-components).

**The goal** is to start shipping rules to enforce best practices in user land while using our icons offering

This is a **scaffold-only** package — it ships a valid plugin skeleton with **no rules yet**, ready to be extended.

## Details

- **Name:** `@fluentui/eslint-plugin-react-icons`
- **Initial state:** `private: true` (not published yet)
- **Release group:** `standalone` (registered in `nx.json`)
- Mirrors the repo's standalone-package conventions (`react-icons-atomic-webpack-loader`): `tsc` build → `lib/`, Vitest, `project.json` with the `npm:standalone` tag.
- Exports an empty `rules` registry plus `recommended` (flat) and `recommended-legacy` (classic) configs.

## Files

- `packages/eslint-plugin-react-icons/` — `package.json`, `project.json`, `tsconfig.json`, `README.md`, `src/index.ts`, `src/rules/index.ts`, `src/index.test.ts`
- `nx.json` — adds the package to `release.groups.standalone.projects`
- `package-lock.json` — registers the new workspace; the `"dev": true` removals reflect `@typescript-eslint/utils` becoming a runtime dependency of a workspace package

## Safety / release verification

Because the package is `private: true`, adding it to the `standalone` release group is safe:

- `nx release version --group=standalone --dry-run` → exits 0, bumps the version while preserving `private: true`.
- `nx release publish --group=standalone --dry-run` → runs for **only the 3 public packages**; the private package is **automatically excluded**.
- The `publish-standalone.yml` workflow additionally filters it out via `--exclude "tag:npm:private"`.

## Verification

- `nx run eslint-plugin-react-icons:build` → produces `lib/` (`.js` + `.d.ts`)
- `nx run eslint-plugin-react-icons:test` → Vitest passes (3 tests)
- Built CJS output loads correctly as an ESLint plugin (`meta` / `configs` / `rules` shape valid)

## Notes

- No changes to the publish workflows are required yet — the package stays private until the first rule is ready and `private: true` is removed.

### DIFF
--- a/packages/eslint-plugin-react-icons/eslint.config.mjs
+++ b/packages/eslint-plugin-react-icons/eslint.config.mjs
@@ -1,0 +1,15 @@
+// @ts-check
+import tseslint from 'typescript-eslint';
+
+export default tseslint.config(
+  {
+    ignores: ['lib/', 'node_modules/', '*.js', '*.mjs'],
+  },
+  ...tseslint.configs.recommended,
+  {
+    files: ['**/*.test.ts'],
+    rules: {
+      '@typescript-eslint/no-explicit-any': 'off',
+    },
+  },
+);
--- a/packages/eslint-plugin-react-icons/package.json
+++ b/packages/eslint-plugin-react-icons/package.json
@@ -1,0 +1,39 @@
+{
+  "name": "@fluentui/eslint-plugin-react-icons",
+  "version": "0.0.0",
+  "private": true,
+  "description": "ESLint plugin for @fluentui/react-icons.",
+  "main": "lib/index.js",
+  "typings": "lib/index.d.ts",
+  "scripts": {
+    "build": "tsc -p .",
+    "lint": "eslint src",
+    "test": "vitest run"
+  },
+  "engines": {
+    "node": ">=20.0.0"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/microsoft/fluentui-system-icons.git",
+    "directory": "packages/eslint-plugin-react-icons"
+  },
+  "license": "MIT",
+  "bugs": {
+    "url": "https://github.com/microsoft/fluentui-system-icons/issues"
+  },
+  "dependencies": {
+    "@typescript-eslint/utils": "7.18.0"
+  },
+  "devDependencies": {
+    "eslint": "8.57.1",
+    "typescript": "5.0.4",
+    "@types/node": "22"
+  },
+  "peerDependencies": {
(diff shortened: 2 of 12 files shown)


packages/eslint-plugin-react-icons/eslint.config.mjs
```diff
@@ -0,0 +1,15 @@
+// @ts-check
+import tseslint from 'typescript-eslint';
+
+export default tseslint.config(
+  {
```
mainframev, line 5: I think for a top level ignores (project ignores) we can use [globalIgnores](https://eslint.org/docs/latest/use/configure/ignore)

and if I am not mistaken mode_modules are ignored by default, so having it explicitly ignored might be redundant

packages/eslint-plugin-react-icons/src/index.ts
```diff
@@ -0,0 +1,36 @@
...
+ * and surfaced through the `recommended` config below.
+ */
+const plugin = {
+  meta,
+  rules,
+  configs: {} as Record<string, TSESLint.FlatConfig.Config | TSESLint.ClassicConfig.Config>,
+};
+
+// Flat config (ESLint 9+, recommended).
+plugin.configs.recommended = {
```
mainframev, line 22: just a heads up that we need to bump ts-eslint in a repo, it still supports only Eslint 8

packages/eslint-plugin-react-icons/package.json
```diff
@@ -0,0 +1,49 @@
+{
+  "name": "@fluentui/eslint-plugin-react-icons",
+  "version": "0.0.0",
+  "private": true,
+  "description": "ESLint plugin for @fluentui/react-icons and @fluentui/react-brand-icons.",
```
mainframev:
```suggestion
  "description": "ESLint plugin for @fluentui/react-icons and @fluentui/react-brand-icons",
```

mainframev (approved):

packages/eslint-plugin-react-icons/src/index.ts
Hotell: yup

PR #1165, opened 2026-07-15
feat(eslint-plugin-react-icons): add prefer-resizable rule

Hotell:
## Summary

Adds the first rule to `@fluentui/eslint-plugin-react-icons`: **`prefer-resizable`**.

It flags usage of fixed-size ("sized") Fluent icons (e.g. `Send24Regular`, `AccessTime24Filled`) imported from `@fluentui/react-icons` (and `@fluentui/react-brand-icons`) and offers an editor **suggestion** to switch to the resizable variant (e.g. `SendRegular`). Resizable icons scale to any size (via `fontSize`/CSS) and reduce bundle size.

> Stacks on #1164 (plugin scaffolding). Until that merges, this PR's diff includes those scaffolding commits.

## How sized icons are detected

Detection is **data-driven and generated from `@fluentui/react-icons` `metadata.json`** (the authoritative `sized` | `resizable` classification), so the runtime rule stays a single regex + `Set` lookup:

- A name is "sized" when it ends with a real two-digit Fluent size (`10,12,16,20,24,28,32,48`) immediately before its variant suffix.
- Names whose trailing digits are part of a product name (e.g. `Battery10Regular`, `Fps120Regular`, `Timer10Filled`) are excluded via a generated **collision denylist** — so genuinely-sized `Presence*10` icons are still flagged while `Battery10` etc. are not.
- `scripts/generate-icon-sizes.mjs` emits `icon-sizes.generated.ts` and **verifies** it classifies every icon exactly as `metadata.json` does (fails if the icon set drifts).

## Coverage

- Named imports (incl. aliases) and subpath imports (`.../fonts/*`, `.../svg/*`).
- Namespace usage: `import * as Icons` → `<Icons.Send24Regular />` and value access.
- Suggestion-only (no auto-fix), since swapping sized→resizable changes optical sizing.

Enabled as `warn` in the `recommended` config.

## Testing

- `nx run eslint-plugin-react-icons:test` — 20 tests (rule + data drift guard).
- `type-check`, `type-check:infra`, `build`, `lint` all green.


packages/eslint-plugin-react-icons/docs/rules/prefer-resizable.md
```diff
@@ -0,0 +1,81 @@
...
+Detection is data-driven, generated from `@fluentui/react-icons` `metadata.json`:
+
+- A name is "sized" when it ends with a real two-digit Fluent size
+  (`10, 12, 16, 20, 24, 28, 32, 48`) immediately before its variant suffix
+  (`Filled` / `Regular` / `Color` / `Light`).
+- Names whose trailing digits are part of a product name rather than a size
+  (e.g. `Battery10Regular`, `Fps120Regular`, `Timer10Filled`) are excluded via a
+  generated collision denylist, so they are never mis-flagged.
+
+The generated data (`src/data/icon-sizes.generated.ts`) is verified to classify
```
mainframev: is this path correct? the file is stored under *src/rules/prfer-resizable/icon-sizes.generated.ts*

mainframev (approved):

packages/eslint-plugin-react-icons/docs/rules/prefer-resizable.md
Hotell: it was in data initially then I refactor, this is wrong. ty

PR #1189, opened 2026-07-28
chore: migrate the repo to yarn 4

Hotell:
## Summary

Migrates the repo from npm workspaces to **Yarn 4.17.1** (corepack managed, `nodeLinker: node-modules`, no PnP, no zero-installs).

Nine atomic commits, each independently cherry-pickable. Commit 1 is a pure npm-side cleanup and commit 9 is a standalone guard — both can be extracted into separate PRs if preferred.

## Why

- npm's lockfile keeps getting regenerated against contributor-local mirrors and platform-pruned installs, producing enormous misleading diffs and recurring CI failures (`Cannot find native binding`, `WorkspaceContext is not a constructor` — [npm/cli#4828](https://github.com/npm/cli/issues/4828)). `yarn.lock` records **all** platform variants with explicit `conditions: os=linux & cpu=x64 & libc=glibc`, so that class of bug cannot recur and the root `optionalDependencies` workaround is deleted.
- Dependency build scripts become **opt-in** (`dependenciesMeta.<pkg>.built`), a meaningful supply-chain improvement. Only `@swc/core`, `esbuild`, `nx` and `ttf2woff2` are enabled.
- `patch-package` is replaced by yarn's native `patch:` protocol, dropping a dependency and a `postinstall` step.

## Commits

| | |
|---|---|
| `chore: inline pre/post npm script hooks` | Yarn has no arbitrary `pre*`/`post*` hooks. Inlined while still on npm, so the change is verifiable in isolation. |
| `chore: bootstrap yarn 4 as the repo package manager` | `.yarnrc.yml`, `packageManager`, `.gitignore`, drop `package-lock.json` + `importer/.npmrc`, add `yarn.lock`. |
| `chore: migrate patch-package to yarn's patch: protocol` | `patches/` → `.yarn/patches/`, docs moved to `docs/patches.md`. |
| `chore: absorb the dependency drift from regenerating the lockfile` | See below. |
| `chore: run workspace binaries through yarn` | 16 files. |
| `ci: install and run through yarn` | 13 workflow/pipeline files. |
| `docs: describe the yarn workflow` | |
| `chore: exempt yarn resolutions from the single version policy` | syncpack `versionGroup`. |
| `chore: guard the package script conventions` | New `yarn scripts:check` CI gate. |
| `chore: declare avocado through a yarn catalog` | Lets `importer` own the binary while `.yarnrc.yml` stays the single source of the version. |
| `refactor: align the package script guard with the repo CLI conventions` | One `main()`, `node:util` `parseArgs`, `--help` / `--verbose`. |
| `fix: hold griffel at the versions the npm lockfile resolved` | Keeps the migration bundle-size neutral. See below. |
| `chore: sweep the remaining npm references` | CODEOWNERS, `.prettierignore`, `.npmrc`, the commented-out flutter/android jobs, `SPEC.md`, instructions and generated-file banners. |

## The big constraint: yarn does not put `node_modules/.bin` on `PATH`

A workspace script may only invoke a binary bare when the package providing it is one of **that workspace's own** dependencies. This repo's [single version policy](docs/single-version-policy.md) keeps every tool in the root manifest, so all 41 bare invocations across 13 packages became `yarn run -T <bin>`.

Three related gotchas, all now documented and enforced:

- The flag only exists on the explicit `run` command — `yarn -T <bin>` is a syntax error.
- **Hot loops get a catalog instead.** `importer`'s avocado loop runs `find -exec ... \;` once per asset directory (~2970), where `yarn run -T` (~438ms) against the bare binary (~112ms) burned ~16 minutes of pure startup. Yarn only puts a workspace's *own* deps on the script `PATH`, so avocado is declared in `importer` with its version held in a root `catalog:` entry -- single version policy intact, bare-name call restored. `-exec ... {} +` batches into one process, so plain `yarn run -T` stays fine there.
- **Yarn runs scripts through its own portable shell, which glob-expands environment assignments.** `SVG_SPRITE_MERGED_FILENAME=fluentui-react-icons.[contenthash].svg` dies with `No matches found`. Quote the value.

### Measured overhead (20 iterations)

| invocation | total | per call |
|---|---|---|
| bare binary on the workspace `PATH` | 0.049s | ~2ms |
| `yarn run -T <bin>` | 3.749s | ~187ms |
| `npx <bin>` (what it replaced) | 8.893s | ~445ms |

`yarn run -T` is ~2.4× **faster** than the `npx` it replaced. Static upper bound across every script in the repo is 104 spawns ≈ 20s, so this is not a hot path anywhere outside the loops noted above.

## Dependency drift — please read

**Yarn 4 cannot import an npm lockfile.** There is no `yarn import` command, and `--mode=update-lockfile` ignores `package-lock.json` entirely (verified with an isolated fixture: an npm-pinned `lodash@4.17.15` resolved to `4.18.1`). Generating `yarn.lock` therefore floats every dependency to the newest version in its declared range. Four of those bumps break the build. Two are fixed at the source; two need an install-time override, both scoped so they hold back only the dependent that needs it.

### Fixed properly

| dependency | drift | failure | fix |
|---|---|---|---|
| `ts-loader` | 9.5.4 → 9.6.2 | changed tsconfig discovery, so the atomic loader test fixtures started resolving the package's own tsconfig → `TS6059: not under rootDir` | the fixtures now declare the `rootDir` they actually sit in |
| `@fluentui/babel-preset-storybook-full-source` | 0.1.1 → 0.1.2 | reads an `importMappings` option its caller `react-storybook-addon-export-to-sandbox@0.2.1` never passes → `Object.keys(undefined)`, all 12 docsite stories fail | upgrade the addon to 0.3.0, which exposes the option, and pass `importMappings: {}` — we publish no unstable entry points needing rewrite on sandbox export |

### Overridden via root `resolutions`

| dependency | drift | failure |
|---|---|---|
| `@griffel/react` | 1.6.1 → 1.7.6 | `.d.ts` uses `import { type X }` (TS 4.5+); `react-icons` builds with a pinned TS 4.1.6 → `TS1005`. Also +1.1 kB minified on the `FileTypeIcon` bundle-size fixture → monosize threshold breach |
| `@griffel/core` | 1.20.1 → 1.21.3 | pulled in with the above; part of the same bundle-size delta |
| `@linaria/shaker/@babel/plugin-transform-runtime` | 7.29.0 → 8.0.1 | `@linaria/shaker` declares an unbounded `">=7"`; `useESModules` was removed in 8 |

The `@linaria/shaker` key is a **scoped `parent/child` selector**, so it holds nothing else back. Griffel is pinned repo-wide on purpose: every workspace that uses it declares `^1.6.1`, and npm's lockfile resolved 1.6.1 / 1.20.1 everywhere, so a bare key is what actually reproduces `main`. Letting the fresh yarn resolve win moved shipped bytes — `FileTypeIcon` grew 1.136 kB minified and failed the 1 kB `monosize` threshold. A package manager migration should not do that. The ranges the packages **publish** stay `^1.6.1`, so consumers still deduplicate griffel with the rest of their app.

`skipLibCheck` was ruled out — `TS1005` is a parse error raised before any checking happens — as was narrowing the published range, which would force every consumer onto exactly 1.6.1. The griffel override disappears once the packages move off TypeScript 4.1.6 and the bundle cost has been re-measured; it is documented in [docs/single-version-policy.md](docs/single-version-policy.md).

## New CI gate: `yarn scripts:check`

[`scripts/check-package-scripts.mjs`](scripts/check-package-scripts.mjs) flags, in every workspace manifest:

- a root-owned binary invoked bare, including after `find -exec` / `xargs`
- `yarn bin <root-bin>`, which silently cannot see root binaries
- `pre*`/`post*` keys outside the lifecycle names yarn implements
- `npx`, `npm run|ci|install|exec`, hardcoded `node_modules/.bin` paths

It follows the repo's CLI conventions (single `main()`, `parseArgs`, `--help`); `--verbose` prints the resolved root/workspace binaries and the tokens it treats as executed programs, which is what you want when a result looks wrong.

Implemented as a script rather than an nx workspace ESLint rule because `svg-icons`, `svg-sprites`, `icon-app` and `importer` have no eslint config and no lint target — and those are exactly the packages whose scripts contain the most shell. All six violation classes are verified to fail the check.

## Release pipeline

`nx release publish` is package-manager agnostic (it shells out to `npm publish` / `npm view` / `npm dist-tag`), so auth stays `.npmrc` + `NODE_AUTH_TOKEN` from `setup-node`'s `registry-url`. The **`npm install --package-lock-only` lockfile refresh steps are deleted** from all three publish workflows: every cross-workspace dependency is `"*"` and yarn records workspace entries as `0.0.0-use.local`, so `nx release version` cannot dirty `yarn.lock`.

Azure pipelines move from Node 18.x to 22.x, and the legacy Node 11.x duplicate under `importer/.azure/` is removed.

## Validation

Everything below was run locally on macOS / Node 22.21.1.

- ✅ `yarn nx run-many -t build,lint,test,type-check:infra,build-verify --skip-nx-cache` — **all 11 projects green**, exit 0 in **9:39**. For reference the npm baseline for `build` alone was 9:00, so this superset of targets lands at roughly parity.
- ✅ **No generated-output drift** — `git status` clean after a full uncached build.
- ✅ `yarn install --immutable` passes (what CI runs). Cold install 30.7s / 684 packages, warm 22.7s, incremental ~4.5s.
- ✅ **Font pipeline is byte-for-byte deterministic** across two runs, and the codepoint JSONs are identical to `main` — the fantasticon `patch:` migration reproduces `patch-package`'s output exactly. (The `.ttf`/`.woff` binaries differ from the committed ones by ~4 bytes, because those are built by CI on Linux; not regenerated here.)
- ✅ `yarn nx release version 1.2.3 --group=native --dry-run` leaves `yarn.lock` untouched; `yarn nx release publish --group=native --access public --dry-run` succeeds.
- ✅ `yarn nx format:check`, `yarn deps:check`, `yarn scripts:check` all clean.
- ✅ Dependency nesting verified: `packages/react-icons/node_modules/typescript` = 4.1.6 while root is 5.0.4, and `yarn tsc` vs `yarn run -T tsc` resolve to each respectively.

## Notes for reviewers

- **`yarn nx reset` is required after any lockfile change**, otherwise the daemon's stale graph throws `Target project does not exist: npm:<pkg>@<version>`. Worth mentioning in the merge announcement.
- Consumer-facing `npm install @fluentui/...` snippets in READMEs and release output are intentionally left as-is.

## Out of scope

Yarn PnP · zero-installs / committing `.yarn/cache` · a composite setup action · changes to the nx release strategy, groups or versioning.


.github/workflows/deploy-docsite.yml
```diff
@@ -42,17 +42,20 @@ jobs:
...
+
       - uses: actions/setup-node@v7
         with:
           node-version: 22.x
+          cache: 'yarn'
 
-      - run: npm ci
-      - run: npx playwright install --with-deps
+      - run: yarn install --immutable
+      - run: yarn playwright install --with-deps
```
mainframev, line 53: blocking: I feel that it won't work without npx, we do not have playwright specified as dependency in package.json

README.md
```diff
@@ -82,23 +82,25 @@ Inline svg directly. See [packages/svg-icons/README.md](packages/svg-icons/READM
 
 The importer generates the Android and iOS libraries from the icons in the `assets` directory.
 
-Jump into the directory:
+Install dependencies from the repository root (this is a single yarn workspace, so
+dependencies for every package are installed in one go):
 
 ```
-cd importer
+corepack enable
```
mainframev, line 89: q: AFAIK corepack will be removed from node >= 25. I know currently engines specified 22 || 24, but maybe we should recommend install corepack globally

mainframev (changes requested):
LGTM, except one blocking about playwright 

there are also files that are not in the diff, but worth to adjust in this PR or in a follow-up: 

- CODEOWNERS package-lock.json should be replaced with yarn.lock 
- prettierignore with package-lock.json
- instead of npx -> yarn dlx? (workflows/pr.yml L240: npx http-server). 
- [SPEC.md](https://github.com/Hotell/fluentui-system-icons/blob/feat/migrate-to-yarn4/packages/react-icons-svg-sprite-subsetting-webpack-plugin/SPEC.md) mentions npm commands
- disabled workflow in PR.yml mentions *npm run deploy:flutter*

README.md
Hotell: yeah it will be removed, lets address is in follow up

.github/workflows/deploy-docsite.yml
Hotell: pw is transitive dep of llm extractor so this never fetched pw from registry 
image

Hotell:
> LGTM, except one blocking about playwright
> 
> there are also files that are not in the diff, but worth to adjust in this PR or in a follow-up:
> 
> * CODEOWNERS package-lock.json should be replaced with yarn.lock
> * prettierignore with package-lock.json
> * instead of npx -> yarn dlx? (workflows/pr.yml L240: npx http-server).
> * [SPEC.md](https://github.com/Hotell/fluentui-system-icons/blob/feat/migrate-to-yarn4/packages/react-icons-svg-sprite-subsetting-webpack-plugin/SPEC.md) mentions npm commands
> * disabled workflow in PR.yml mentions _npm run deploy:flutter_

valid, lemme address those ty!

.github/workflows/deploy-docsite.yml
Hotell: update: so that transitive works for npx, not for yarn which is strict about resolution -> adding pw to root devDeps. ty!

mainframev (approved):

PR #1196, opened 2026-07-30
chore: bump fluent storybook addons + adopt shared preview styles

Hotell:
## Description

Updates `@fluentui/react-storybook-addon` to `0.7.1` and `@fluentui/react-storybook-addon-export-to-sandbox` to `0.3.1`, and adopts what those releases changed.

### 1. Yarn refused to install them

```
YN0016: @fluentui/react-storybook-addon-export-to-sandbox@npm:0.3.1:
        All versions satisfying "0.3.1" are quarantined
```

Yarn 4.10 added [`npmMinimalAgeGate`](https://yarnpkg.com/configuration/yarnrc#npmMinimalAgeGate), which refuses to resolve versions published more recently than the gate (default `3d`) as a supply-chain mitigation. npm has no equivalent check, which is why the same install succeeds there.

The gate is worth keeping for the ecosystem at large, but it blocks us from consuming our own packages on release day. So first-party `@fluentui/*` is exempted via `npmPreapprovedPackages` rather than lowering the gate repo-wide — everything else still has to age.

### 2. Docsite: shared preview styles

`react-storybook-addon` 0.7.0 upstreamed the docs preview stylesheet ([microsoft/fluentui#35949](https://github.com/microsoft/fluentui/pull/35949)), which is exactly what the `TODO` in `preview.tsx` was waiting on. It ships as `@fluentui/react-storybook-addon/styles.css` and, per the README, must be imported by the consumer — the preset does not inject it.

- `docs-root.css` (478 lines) verified rule-by-rule to be a strict subset of the published stylesheet → deleted. The one behavioural delta is in our favour: upstream drops `padding: 0` from `.sbdocs-preview`, i.e. the Canvas Toolbar fix from [microsoft/fluentui#36223](https://github.com/microsoft/fluentui/pull/36223) that was never backported here.
- `docs-root-v9.css` was upstreamed too, except the `.fluent-version-badge` rules backing a docsite-local component. Those survive in `docsite.css`, renamed so the name describes what it holds rather than a storybook version we no longer inherit from.

Net: **-505 lines** of vendored CSS.

### 3. Docsite: exported sandboxes ran the wrong React

Found while auditing the `export-to-sandbox` bump. `parameters.exportToSandbox.requiredDependencies` pinned `react`/`react-dom` to `^18` while the repo and its docs build against 19, so every exported StackBlitz ran a different React than the story it came from.

The addon's own `commonDevDeps` are staler still — `@types/react@^17`, `@types/react-dom@^17`, `typescript@~4.7.0`. They're spread before `parameters…devDependencies`, so overriding them here is enough. All three move together: the generated `package.json` runs `tsc && vite build`, and React 19 types don't compile under TS 4.7.

`export-to-sandbox` 0.3.1 itself needed no config change — its only change ([microsoft/fluentui#36385](https://github.com/microsoft/fluentui/pull/36385)) moves the babel rule to `enforce: 'pre'` so the story-source transform sees original source. Verified the transform still runs under `addon-webpack5-compiler-swc`: `fullSource` is injected into 19 story bundles.

## Verification

- `yarn install` resolves cleanly, no `YN0016`.
- `yarn nx run docsite:build` passes; shared + docsite-only CSS and the corrected sandbox dependency map are all present in the emitted bundle.
- `yarn nx run docsite:lint` passes.


.yarnrc.yml
```diff
@@ -2,6 +2,13 @@ nodeLinker: node-modules
 
 npmRegistryServer: 'https://registry.npmjs.org'
 
+# Yarn quarantines versions published more recently than `npmMinimalAgeGate`
+# (default 3d) as a supply-chain mitigation, failing installs with YN0016.
+# First-party Fluent UI packages are exempt so we can consume them the day they
+# ship; keep the gate on for everything else.
+npmPreapprovedPackages:
+  - '@fluentui/*'
```
mainframev, line 10: q: should we add here griffel/* packages too?

mainframev (approved):

PR #1199, opened 2026-07-31
feat(react-icons): add opt-in native ESM build behind --enable-native-esm

Hotell:
## Summary

Lands the tooling that turns the dual build into valid **bare-Node ESM**, without changing anything the package ships.

`node scripts/build.js --enable-native-esm` is **off by default**. With the flag off, `lib/` and `lib-cjs/` are byte-for-byte identical to `main` and the manifest keeps its current `type`/`main`/`exports`. A follow-up PR flips the default.

Landing it now — rather than parking it on a branch until the format is switched on — means the codemod is exercised on **every icon change** instead of silently drifting away from the generated output. That drift is the actual risk: the codemod rewrites ~23k emitted files derived from `assets/`.

## What the flag does

| | default (off) | `--enable-native-esm` |
|---|---|---|
| `lib/` specifiers | extensionless | fully specified (`./x.js`, `/index.js`) |
| `lib-cjs/` | `.js` + `.d.ts` | `.cjs` + `.d.cts` |
| manifest | untouched | `type: module`, `.cjs` main, `import`/`require` exports |

- **`fullySpecifyEsm`** — Node's ESM resolver requires extensions; tsc 4.x cannot emit them.
- **`finalizeCjs`** — renames `lib-cjs/` output, required once the package is `type: module`, where a bare `.js` would be read as ESM.
- **`applyEsmFirstManifest`** — flips the manifest to match. It transforms *whatever is already in* `exports`, so the subpaths appended at build time (sprite, headless) convert too and their emitters stay format-agnostic. Verified to reproduce the previously hand-written ESM-first map **exactly**.

## Two non-obvious constraints (both encoded in the code)

1. **The conversion runs last in the build.** Rewriting the manifest to `type: module` changes how Node interprets every `.js` file in the package, so nothing may be `require`d after it.
2. **`scripts/package.json` pins the build scripts to CommonJS.** A no-op today; without it a *second* `--enable-native-esm` run fails, because the build scripts get re-read as ESM. (This was one of the "hygiene bits" originally deferred — it turns out to be load-bearing for the flag to be re-runnable.)

## Verification

Real e2e, not a resolver simulation: `verify-esm-exports.{mjs,cjs}` derive every specifier from `exports` and load them in **spawned Node processes**, covering the `import`/`require` conditions and `fluentIconFont` routing.

Font and `.css` entries are resolve-only — font modules pull in binary `.ttf`/`.woff` assets that only a bundler can load. The smoke test **skips unless the build is `type: module`**: a bare-Node `import` failing in the dual build is the limitation being removed, not a regression.

Evidence from this branch:

- **Acceptance gate — flag off ⇒ zero change:** `diff -r` vs a pristine `main` build → `lib/` and `lib-cjs/` **identical**; manifest diff is scripts-only.
- `build-verify.test.js` has **zero diff** vs `main` (135 passed / 12 skipped), so the default path's assertions are provably unweakened.
- Flag on: build + all 4 e2e checks green (`esm · default`, `esm · fluentIconFont`, `cjs · default`, `cjs · fluentIconFont`); re-runnable.
- `lint`, `type-check:infra`, `scripts:check`, unit tests (90) all pass.

## CI

New `verify-native-esm` job runs `nx affected -t verify-native-esm`, so it triggers whenever react-icons or the icon assets change. It's a **separate job on purpose**: the native ESM build rewrites `packages/react-icons/package.json` in place, which would otherwise leak into the packaging and bundle-size steps of `build-npm`.

## Follow-up

A subsequent PR flips the default and deletes the flag (existing draft #1143 carries the manifest work).

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -161,6 +161,42 @@ jobs:
   #       accessToken: "placeholder"
   #       refreshToken: "placeholder"
 
+  verify-native-esm:
+    if: ${{ github.repository_owner == 'microsoft' }}
+    name: Verify native ESM output
+    runs-on: ubuntu-latest
+    permissions:
+      contents: 'read'
+      actions: 'read'
+
+    steps:
+      - uses: actions/checkout@v7
+        with:
+          fetch-depth: 0
+          filter: tree:0
+
+      - run: corepack enable
+
+      - uses: actions/setup-node@v7
+        with:
+          node-version: 22.x
+          cache: 'yarn'
+
+      - run: yarn install --immutable
+
+      - uses: nrwl/nx-set-shas@afb73a62d26e41464e9254689e1fd6122ee683c1 # v5.0.1
+
+      # `@fluentui/react-icons` is still published as the historical dual build; native
+      # ESM-first output is opt-in behind `build.js --enable-native-esm`. Building that
+      # variant here keeps it from drifting as icons are added or changed, so flipping
+      # the default later stays a one-line change rather than a rediscovery exercise.
+      #
+      # This runs in its own job on purpose: the native ESM build rewrites
+      # `packages/react-icons/package.json` in place (type/main/exports), which would
+      # otherwise leak into the packaging and bundle-size steps of `build-npm`.
+      - name: Build and verify native ESM output
+        run: yarn nx affected -t verify-native-esm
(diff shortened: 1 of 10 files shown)


mainframev (approved):
LGTM

PR #1205, opened 2026-08-07
feat(react-icons): ship native ESM output by default

Hotell:
## Summary

Makes the previously opt-in `--enable-native-esm` build the permanent, shipped default for `@fluentui/react-icons`.

`@fluentui/react-icons` is now **ESM-first**:

- the package is marked `"type": "module"`
- `lib/` is native ESM — every relative specifier is fully specified (`./foo.js`)
- `lib-cjs/` is finalized CommonJS — `.cjs` / `.d.cts`
- consumers are routed through nested conditional `exports` (`import` / `require`, each with its own `types`)

This is **not** a breaking change:

- entry points, subpath exports and type resolution are unchanged
- bundlers (webpack/vite/rollup/metro) resolve the exact same ESM they already resolved
- `require()` keeps working via `lib-cjs/`
- `module` and `typings` are retained for legacy tooling
- `bundler` module resolution stays the recommendation; `node16` is **not** required

Docs state the first shipping version as `v2.0.335`.

## Key change: the manifest is committed, not mutated

The opt-in implementation rewrote `package.json` in place at the end of the build (`applyEsmFirstManifest`). That was safe only while it ran last, and it poisoned the working tree for any subsequent build.

The ESM-first manifest is now **committed as the source of truth** and the build never mutates `type` / `main` / `exports`. The sprite and headless export emitters produce nested conditions via a shared `esmFirstEntry()` helper instead of flat strings.

### `@fluentui/react-icons`

- `package.json` — `"type": "module"`, `main: lib-cjs/index.cjs`, all 18 subpath exports nested
- `scripts/build.js` — flag + `convertToNativeEsm()` removed; `fullySpecifyEsm()` / `finalizeCjs()` run last on purpose
- `scripts/module-format.js` — `applyEsmFirstManifest` and friends dropped; `esmFirstEntry()` added
- renamed package-level CJS tooling: `svgo.config.js` → `.cjs`, `fluent-icons-transform.js` → `.cjs`
  (`scripts/*.js` intentionally untouched — `scripts/package.json` declares `{"type":"commonjs"}`)
- new `check:types-packaging` target running `@arethetypeswrong/cli`; wired into the `build-npm` PR job
- removed the `build-native-esm` / `verify-native-esm` targets and the dedicated `build-npm-native-esm` CI job
- `build-verify` assertions + snapshots updated for the `.cjs` / `.d.cts` shape; two previously vacuous chunk-structure loops now assert they actually iterate
- README + docsite `Module Formats` section

### Build tooling

- `react-icons-font-subsetting-webpack-plugin` and `react-icons-svg-sprite-subsetting-webpack-plugin`: their module matchers already had a `lib(-cjs)?` branch but required a `.js` suffix, so that branch went dead once `lib-cjs/` started emitting `.cjs`. Patterns now accept `.c?js`. The `lib/` path — what bundlers actually resolve for ESM app code — was unaffected.
- `react-icons-atomic-webpack-loader` needs no change: it emits bare specifiers (`@fluentui/react-icons/svg/<icon>`) and lets the resolver pick the condition.

## Validation

- full clean build green — `fullySpecifyEsm` 23399/23422 files, `finalizeCjs` 11713 js→cjs / 11709 d.ts→d.cts
- **two back-to-back builds leave `git diff packages/react-icons/package.json` empty** (the mutation hazard is gone)
- bare-Node verifiers pass: `verify-esm-exports.mjs` and `.cjs`, each with and without `--conditions=fluentIconFont`
- `build-verify`: 139 passed / 8 skipped (sprite tests are opt-in)
- `attw --profile node16` → exit 0 (node16-CJS, node16-ESM, bundler all green)
- font subsetting e2e green: every entry in the plugin's webpack fixture stays under its 1.5–3 KB subset-font threshold (unsubset fonts are 100 KB+), including the barrel → atomic-loader → headless-fonts end-to-end entry
- svg sprite subsetting e2e green in both `atomic` and `merged` modes
- `nx run-many -t build,lint,test` green for `icon-app`, `docsite` and all three webpack plugins/loaders
- `monosize` bundle-size runs clean against the existing baseline
- `npm pack --dry-run`: 46886 files, 11713 `.cjs`, 11709 `.d.cts`, **0** stray `lib-cjs/**/*.js` or `*.d.ts`

## Notes for reviewers

- `attw` runs with `--profile node16` and excludes the two `.css` entrypoints (they are not JavaScript). The default `strict` profile fails `node10` for every subpath, which is expected — `node10` predates `exports` maps.
- `publish.yml` needs no change: the build runs before `nx release version` and the manifest is no longer generated.


packages/react-icons/README.md
```diff
@@ -285,6 +285,22 @@ These are the only two types of breaking changes that deviate from standard sema
 
 **Recommendation**: When upgrading, even for patch versions, review the release notes for any removed or renamed icons to ensure your application isn't affected.
 
+### Module Formats
+
+> **Since `v2.0.335`.**
```
mainframev: is this correct? I see v2.0.335 is already shipped

packages/docsite/stories/Icons/IconsAtomicAPI.mdx
```diff
@@ -64,6 +64,14 @@ Without this setting, TypeScript will not be able to resolve the individual icon
 
 **Note:** TypeScript users that are still using old `"moduleResolution": "node"` can use `@fluentui/react-icons/lib/atoms/svg/*` for SVG-based atomic icon imports.
 
+#### Module Formats
+
+**Since `v2.0.335`.**
```
mainframev: same questions as above, if not correct -> update, ignore if it's ok

mainframev (approved):

PR #1212, opened 2026-08-11
feat(svg-sprites): add grouped sprite bundles

codysorgenfrey:
Adds grouped sprite bundles by size/style while keeping the existing per-icon sprite output.

The user can then bundle them as static assets and use any icon they'd like without having to install each icon:
```html
<svg>
  <use href="/assets/fluent_icons_20_regular.sprites.svg#add" />
</svg>
```

This change creates files like `fluent_icons_20_regular.sprite.svg` and `fluent_icons_20_filled.sprite.svg` with ids like `access_time`, `add`, and `alert` so the bundle can be used as a single symbols sheet per size/style without repeating the suffix in every id.

It also updates the sprite build verification to accept the new bundle naming format and documents the grouped bundle output in the SVG sprites README.

### DIFF
--- a/packages/svg-sprites/README.md
+++ b/packages/svg-sprites/README.md
@@ -42,8 +42,8 @@ export const Icon = (props: IconProps) => {
 The package also emits grouped sprite files following the local bundle format, for example:
 
 ```text
-fluent-regular-20.sprite.svg
-fluent-filled-20.sprite.svg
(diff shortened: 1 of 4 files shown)


codysorgenfrey:
@microsoft-github-policy-service agree company="Microsoft"

PR #1218, opened 2026-08-18
refactor(react-icons): generate font-family definitions from one manifest

Hotell:
> [!NOTE]
> Stacked on #1217 — the first commit here is that fix. Review the second commit, or merge #1217 first and this diff shrinks to it.

## Why

The CSS `font-family`, the font binary base name, the `data-fui-icon-font` value and the `FontFile` enum value were duplicated across **six** hand-maintained places:

| File | Duplicated knowledge |
|---|---|
| `src/utils/fonts/createFluentFontIcon.shared.ts` | `FontFile` enum values |
| `src/utils/fonts/createFluentFontIcon.styles.ts` | 12 asset imports, `@font-face` block, `useRootStyles` variants |
| `src/headless/fonts/styles.css` | `@font-face` block |
| `src/headless/styles.css` | `[data-fui-icon-font='…']` rules |
| `src/headless/fonts/createFluentFontIcon.tsx` | `FONT_VARIANT_MAP` |
| `scripts/copy-base-fonts.js` + `scripts/convert-font.utils.js` | family list, hardcoded `0/1/2/3` style codes |

That is exactly how the Light family ended up declared on the headless surface but emitted as `font-family: undefined` on the Griffel one (#1217) — nothing failed the build, the icons just rendered with no font.

## What

- **`scripts/font-families.js`** is the single source of truth (`key` / enum `value` / `family` / `variant` / `file` / `resizable`).
- **`scripts/generate-font-styles.js`** rewrites the derived `#region @generated <name>` blocks *in place*. The surrounding code stays hand-written, committed and type-checked — only the derived data is generated, so the prose comments (the `overflow: hidden` baseline note, the `:where()` specificity note) survive and the files remain reviewable in a normal diff. `--check` fails instead of writing, for a stale checkout.
- Wired in as `yarn generate:font-styles`, part of `build:fonts-and-svg`.
- `copy-base-fonts.js` and `convert-font{,.utils}.js` now consume the manifest instead of their own family lists and magic style codes.

## Guardrails

- `scripts/generate-font-styles.test.js` — pins the `FontFile` values (they are baked into ~24k generated atoms), asserts unique families/variants/files, asserts **every** `@font-face` has a `font-display`, and asserts the checked-in sources match the generated regions.
- `build-verify.test.js` — new assertion that the **shipped** artifacts agree: the families declared in `lib/utils/fonts/createFluentFontIcon.styles.js` and in `lib/headless/{fonts/,}styles.css` are identical and match the manifest. This is the regression test for #1217.

## Notes for review

- The `@font-face` rules are now emitted in manifest order (Filled, Regular, Resizable, Light); Light used to sit third. Distinct families, so no cascade impact — it is visible in the snapshot diff.
- `convert-font.js` keeps an explicit `SIZED_ORDER = ['Light', 'Filled', 'Regular']`: that order is load-bearing for chunk assignment, so it is deliberately *not* derived from the manifest.
- Griffel's babel preset needs the template literal inline in `makeStaticStyles(...)`, so the generated region spans the whole statement.

## Verification

- `yarn nx run react-icons:build` + `build-verify` — 140 passed / 8 skipped, `metadata.json` unchanged (no chunk churn).
- `yarn test` — 112 passed. `lint`, `type-check:infra`, prettier clean.
- Smoke-tested by renaming a family in the manifest: all four surfaces regenerated together.


packages/react-icons/scripts/font-families.js
```diff
@@ -0,0 +1,85 @@
...
+  woff2: 'woff2',
+  woff: 'woff',
+  ttf: 'truetype',
+});
+
+/**
+ * @param {string} key
+ * @returns {FontFamilyDefinition}
+ */
+function getFontFamilyByKey(key) {
```
Hotell, line 72: this should be string union rather `string`

packages/react-icons/scripts/font-families.js
```diff
@@ -0,0 +1,85 @@
...
+    file: 'FluentSystemIcons-Regular',
+    resizable: false,
+  },
+  {
+    key: 'Resizable',
+    value: 2,
+    family: 'FluentSystemIcons',
+    variant: 'resizable',
+    file: 'FluentSystemIcons-Resizable',
+    resizable: true,
```
Hotell, line 46: what is this `resizable`  about ?

PR #1219, opened 2026-08-20
feat(react-icons-font-subsetting-webpack-plugin): support rspack

Hotell:
## Problem

`@fluentui/react-icons-font-subsetting-webpack-plugin` crashes under rspack:

```
Compilation.hooks.optimizeAssets is not supported in rspack
  at react-icons-font-subsetting-webpack-plugin/lib/index.js:65:25
```

Fixing only the hook is not sufficient. The plugin also gated module selection on
`m instanceof webpack.NormalModule`, which **no rspack module satisfies** — so it would
have run and subset nothing. A silently-useless plugin is worse than a crash, so this PR
addresses the full set of incompatibilities rather than just the reported one.

## Root causes

Six distinct issues, all verified against rspack 2.x:

| Before | rspack behaviour | After |
| --- | --- | --- |
| `import * as webpack from 'webpack'` (runtime import) | fails when only rspack is installed | no bundler imported at all; reached via `compiler.webpack` |
| `compilation.hooks.optimizeAssets` | **not supported** — the reported crash | `hooks.processAssets` at `PROCESS_ASSETS_STAGE_OPTIMIZE` |
| `m instanceof webpack.NormalModule` | never true; rspack modules are napi proxies | duck-typing on `resource` |
| `getUsedExports(m, undefined)` | binding requires `string \| string[]`, not optional | explicit runtime (entrypoint names) under rspack, `undefined` under webpack |
| `m.buildInfo.filename` | rspack leaves `buildInfo` **entirely empty** for asset modules | `AssetInfo.sourceFilename`, populated by both bundlers |
| `compilation.assets[name] = …` | assets record is a read-only proxy; writes never reach Rust | `compilation.updateAsset()` |

`getProvidedExports`, `compilation.modules` and the atomic loader needed no changes.

## Changes

- **`src/index.ts`** — all six fixes above. The compiled output requires only `subset-font`,
  `path` and `fs/promises`; no bundler is imported at runtime.
- **`src/bundler-api.ts`** — the bundler surface the plugin depends on, hand-maintained so that
  neither bundler is imported *even for types*. Every member is deliberately wider than the
  corresponding webpack/rspack type, which keeps real compilers assignable and lets
  `apply(compiler: BundlerCompiler)` be typed rather than `unknown`.
- **Loud failure** — when font modules are found but no asset can be mapped, a warning is
  emitted instead of silently shipping a full un-subset font.
- **`peerDependencies`** — `webpack` and `@rspack/core` are both **optional**, so consumers
  install only the bundler they use. (This also clears a pre-existing yarn peer warning from
  `docsite`.)
- **Tests** — the config factory is shared, running the same five fixture entries against both
  bundlers. `test/run.js` (`--bundler webpack|rspack|all`) replaces the webpack-CLI invocation;
  no `@rspack/cli` dependency needed.
- **`test/types.conformance.ts`** — a `tsc`-only check asserting both bundlers' real `Compiler`
  types still satisfy the maintained interfaces, and that the plugin still satisfies
  `WebpackPluginInstance` and `RspackPluginInstance`. This is the drift protection the
  hand-written types would otherwise lack. Runs as part of `yarn test`.
- **`@rspack/core` is installed in the root `package.json` only**, per the single version policy.

## Review feedback addressed

- **webpack types leaked into `lib/index.d.ts`**, which would have broken type-checking for
  rspack-only consumers — exactly the scenario the optional peers promise. No bundler is
  referenced by the emitted declarations any more; verified by type-checking a consumer in an
  isolated tree with **neither** bundler installed.
- **`as never` on `getUsedExports`** is gone; the maintained `BundlerModuleGraph` declares a
  runtime parameter that covers both bundlers honestly, so no cast is needed.
- **`parseArgs` and older Node** — declined with reasoning: `test/run.js` is dev-only
  (`files: ["lib/*"]`) and contributors are bound by the root manifest's `^22 || ^24`.

## Verification

- `nx run …:build` — clean
- `nx run …:test` — type conformance + webpack ✅ + rspack ✅
- Emitted font sizes are **identical** across bundlers per entry (1.14 KiB / 1 KiB / 892 / 824 / 812 / 772 / 688 / 656 / 632 / 512 / 440 / 320 bytes)
- **Negative test**: removing the plugin from the rspack config makes the suite fail with a
  1,024,976-byte font — proving the tests detect a silent no-op rather than just a crash
- **Conformance negative test**: narrowing any maintained type fails the check
- Consumer type-checks against the published `.d.ts` with neither bundler installed
- `npm pack --dry-run` confirms both `lib/bundler-api.*` files ship
- `nx run …:lint` and `yarn deps:check` — pass

## Follow-up

`@fluentui/react-icons-svg-sprite-subsetting-webpack-plugin` has the same class of problem
but a much larger surface (`RuntimeModule`/`STAGE_ATTACH`, `addRuntimeModule`,
`webpack.util.createHash`, `html-webpack-plugin` `getHooks`, `chunkGraph`). rspack supports
all of these, but it warrants its own PR.


packages/react-icons-font-subsetting-webpack-plugin/src/index.ts
```diff
@@ -1,4 +1,4 @@
-import * as webpack from 'webpack';
+import type * as webpack from 'webpack';
```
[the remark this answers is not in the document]
Hotell: Good catch, and confirmed — the old `lib/index.d.ts` started with `import type * as webpack from 'webpack'`, so an rspack-only consumer would indeed have failed type-checking.

Fixed in a4b8d3d by taking the first option you suggested: the webpack namespace types are replaced with a minimal structural description of the APIs the plugin actually uses, kept module-private so nothing reaches the emitted `.d.ts`.

One wrinkle worth recording: making the *public* signature structural as well broke assignability to webpack's `plugins` array —

```
Type 'FluentUIReactIconsFontSubsettingPlugin' is not assignable to type 'WebpackPluginInstance'.
  ... Type 'Module' has no properties in common with type 'Module'.
  ... 'hooks.compilation.tap' signatures are incompatible
```

So `apply()` now takes `unknown` and narrows once internally. The internals stay precisely typed; only the boundary is permissive.

Verified two ways:

1. Both bundlers still accept the plugin — `tsc` clean on `plugins: [new Plugin()]` typed as `webpack.Configuration` *and* `@rspack/core`'s `Configuration`, plus `apply()` against each real `Compiler`.
2. A consumer type-checks against the shipped `.d.ts` in an isolated tree with **neither** bundler installed — `tsc` clean.

This also removed the `as never` cast you flagged at line 199: the local `ModuleGraph` type declares the runtime parameter as `string[] | undefined`, which covers both bundlers honestly, so no cast is needed.

packages/react-icons-font-subsetting-webpack-plugin/test/run.js
```diff
@@ -0,0 +1,95 @@
...
+      throw new Error(`invalid --bundler "${values.bundler}", expected one of: all, ${BUNDLERS.join(', ')}`);
+    }
+
+    return { bundler: /** @type {'all' | 'webpack' | 'rspack'} */ (values.bundler) };
+  } catch (err) {
+    console.error(`Error parsing arguments: ${/** @type {Error} */ (err).message}`);
+    printUsage();
+    process.exit(1);
+  }
+}
```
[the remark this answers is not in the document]
Hotell: Checked this one, and I'd rather not add a fallback parser.

`test/run.js` is dev-only — the package publishes `files: ["lib/*"]`, so it never reaches consumers. The contributors who run it are bound by the **root** manifest, which is `"node": "^22.0.0 || ^24.0.0"`, and `util.parseArgs` has been stable since Node 20. The repo's toolchain (Nx 22, Vitest 3) already requires Node 22+ independently.

The `>=14.5.0` in this package's `engines` describes the runtime needed to *run the built plugin* (`lib/`), not to run its tests, and that part is still accurate.

So a hand-rolled arg parser would only serve a Node version the repo cannot be developed on. Happy to tighten this package's `engines.node` in a follow-up if the 14.5 floor is considered misleading, but that is a consumer-facing change and orthogonal to rspack support.

packages/react-icons-font-subsetting-webpack-plugin/package.json
```diff
@@ -30,8 +33,17 @@
     "subset-font": "^1.4.0"
   },
   "peerDependencies": {
+    "@rspack/core": ">=1.0.0",
```
mainframev: is peer version correct for rspack? 

`getProvidedExports` (v2.0.0) and `getUsedExports` (2.1.0), both available only > 2.1.0 

https://github.com/web-infra-dev/rspack/releases?page=3#release-v2.0.0
https://github.com/web-infra-dev/rspack/releases?page=2#release-v2.1.0

our root is ^2.1.10

mainframev (approved):

packages/react-icons-font-subsetting-webpack-plugin/package.json
Hotell: very good catch ! with optional meta this doesn't matter much but we should have our requirements tight

Hotell: Good catch — the range was wrong, fixed in f486ea4 as `>=2.1.0`.

One correction to the attribution: the two are the other way round.

| API | Exposed to rspack's JS API | PR |
| --- | --- | --- |
| `moduleGraph.getUsedExports()` | **2.0.0** | [#13519](https://github.com/web-infra-dev/rspack/pull/13519) |
| `moduleGraph.getProvidedExports()` | **2.1.0** | [#14349](https://github.com/web-infra-dev/rspack/pull/14349) |

So `getProvidedExports` is the binding constraint, and your conclusion holds: the plugin needs both, so the floor is 2.1.0.

Worth spelling out why this mattered more than a version-number nit. `getProvidedExports` is not on a rare path — it backs the namespace-import case (`import * as ns from '@fluentui/react-icons/fonts/…'`), which the `atomsImportStar` fixture covers. On rspack 2.0.x the plugin would have worked fine for named imports and then thrown the first time someone wrote a namespace import. A partial failure like that is exactly the kind of thing an over-broad peer range hides.

Also recorded the constraint next to `getProvidedExports` in `bundler-api.ts`, so the peer range doesn't get silently invalidated if that interface is edited later.

Verified after the change: root `^2.1.10` is a strict subset of `>=2.1.0`, both bundler suites still pass, and `yarn install --immutable` is clean (the peer edit did drift the lockfile, which is committed).

On the sibling PR #1224 (atomic loader harness) I left `@rspack/core` at `>=1.0.0`, since that package only touches the loader context (`getOptions` / `callback` / `emitWarning` / `resourcePath`), which has been stable since 1.x. Happy to tighten it to the version actually exercised in CI if you'd rather we only advertise what we test.

PR #1234, opened 2026-09-04
fix(react-icons-font-subsetting-webpack-plugin): detect duplicate installed instances

Hotell:
## Summary

Reported from userland: icons rendering as missing glyphs when several copies of `@fluentui/react-icons` are installed. The plugin was subsetting a shared font for one copy and deleting every glyph the other copies contributed — silently, with only a warning that misdiagnosed the cause.

The plugin now detects this and leaves the affected fonts un-subset, so all glyphs render. `onDuplicateInstances: 'error'` fails the build instead.

## The bug

Attribution runs **asset → `AssetInfo.sourceFilename` → package**, which assumes package ↔ asset is 1:1.

Content hashing breaks that. Copies of the same version ship byte-identical fonts, so they hash to one emitted asset, and that asset can name only one copy as its source. The others match nothing:

```
P3 → shared asset  → subset to P3's icons only
P1, P2, P4, P5 → []  → warned "could not map any font module to an emitted asset"
```

Every glyph reached only through P1/P2/P4/P5 was deleted. The old warning told people to add a `type: 'asset'` module rule, which was not the problem.

Reproduced as a failing unit test before any source change:

```
× leaves fonts whole when a copy owns no emitted asset  → expected [ Array(1) ] to deeply equal []
× fails the build instead when asked to                 → expected [] to have a length of 1
```

That first failure is the production bug: the shared asset *was* updated, i.e. subset down to one copy's icons.

## Why detect rather than compensate

Two approaches were considered and rejected.

**Unioning used exports across copies** (what the reporting app patched locally) works, but only by assuming that copies of the same package name ship identical font content. That is the same class of silent assumption that produced this bug.

**Per-asset unioning** does not work at all: the orphaned copies map to no asset, so there is nothing to union them into.

Detection is scoped to the collision signature — some packages map to assets while others map to none. Copies of genuinely *different* versions ship different fonts, emit separate assets, all map, and continue to be subset independently. The existing all-zero case still reports the missing asset rule, and there is a test pinning that discrimination.

## Options API

The plugin previously took no constructor options:

```js
new FluentUIReactIconsFontSubsettingPlugin({ onDuplicateInstances: 'error' }); // default: 'warn'
```

The warning names both the copies owning an emitted font and the copies sharing them, and points at `resolve.alias` dedupe — with the caveat that collapsing copies also binds them to a single React and Griffel instance, so it is not free advice.

## Conformance gap found along the way

`BundlerCompilation` gained `errors: Error[]`. Verifying that turned up a hole: the conformance test only asserted `webpack.Compiler`/`rspack.Compiler`, so `BundlerCompilation` was reached solely through `hooks.compilation.tap` — whose parameters are **bivariant**, meaning it would have accepted members neither compilation has.

Direct assertions added for both compilations. Confirmed non-vacuous by injecting a bogus member (four errors, including the two new ones).

Manual verification of the widened member:

| bundler | declaration | assignable |
| --- | --- | --- |
| webpack | `Compilation.errors: Error[]` | exact |
| rspack | `get errors(): RspackError[]`, `RspackError extends Error` with all extras optional | both directions |

And at runtime, since types say nothing about a napi-backed array accepting a plain `Error`:

```
webpack: hasErrors=true errorCount=1 warningCount=1
rspack:  hasErrors=true errorCount=1 warningCount=1
```

## Lint rule

Private members now require the `__` prefix. Note this needed the package's lint script widened to `eslint src package.json` — it previously covered only `package.json`, so no rule could have applied to its sources. Verified the rule reports a violation when the prefix is removed. Linting `src` for the first time surfaced no pre-existing problems.

## Validation

- 9/9 unit, type conformance, webpack 6/6, rspack 6/6, lint clean

## Not done

No integration fixture for the duplicate case — reproducing two installed copies in the bundler suite means materialising a second `lib` tree, and the unit harness expresses the condition exactly.

### DIFF
--- a/packages/react-icons-font-subsetting-webpack-plugin/src/index.ts
+++ b/packages/react-icons-font-subsetting-webpack-plugin/src/index.ts
@@ -51,10 +51,10 @@ export interface FluentUIReactIconsFontSubsettingPluginOptions {
 }
 
 export default class FluentUIReactIconsFontSubsettingPlugin implements BundlerPlugin {
-  private readonly onDuplicateInstances: 'warn' | 'error';
+  private readonly __onDuplicateInstances: 'warn' | 'error';
 
   constructor(options: FluentUIReactIconsFontSubsettingPluginOptions = {}) {
-    this.onDuplicateInstances = options.onDuplicateInstances ?? 'warn';
+    this.__onDuplicateInstances = options.onDuplicateInstances ?? 'warn';
   }
 
   /**
@@ -151,7 +151,7 @@ export default class FluentUIReactIconsFontSubsettingPlugin implements BundlerPl
                 `and Griffel instance.`,
             );
 
-            if (this.onDuplicateInstances === 'error') {
+            if (this.__onDuplicateInstances === 'error') {
               compilation.errors.push(message);
             } else {
               compilation.warnings.push(message);
--- a/packages/react-icons-font-subsetting-webpack-plugin/src/index.ts
+++ b/packages/react-icons-font-subsetting-webpack-plugin/src/index.ts
@@ -39,7 +39,24 @@ const UNRESOLVABLE_NAMESPACE_IMPORT = Symbol('unresolvable-namespace-import');
 const REACT_ICONS_FONT_MODULE_IMPORT_PATTERN =
   /react-icons[\/\\]lib(-cjs)?[\/\\](fonts[\/\\](sizedIcons|icons)[\/\\]chunk-\d+|atoms[\/\\](headless-)?fonts[\/\\][\w-]+)\.c?js$/;
 
+export interface FluentUIReactIconsFontSubsettingPluginOptions {
+  /**
+   * What to do when several installed copies of `@fluentui/react-icons` share one emitted font
+   * asset, which makes it impossible to tell which icons belong to it.
+   *
+   * `'warn'` (default) leaves the affected fonts un-subset, so every glyph still renders.
+   * `'error'` fails the build instead.
+   */
+  onDuplicateInstances?: 'warn' | 'error';
+}
+
 export default class FluentUIReactIconsFontSubsettingPlugin implements BundlerPlugin {
+  private readonly onDuplicateInstances: 'warn' | 'error';
+
+  constructor(options: FluentUIReactIconsFontSubsettingPluginOptions = {}) {
+    this.onDuplicateInstances = options.onDuplicateInstances ?? 'warn';
+  }
+
   /**
    * Entry point for the bundler plugin that registers hooks to perform font subsetting for `@fluentui/react-icons`.
    *
@@ -91,6 +108,10 @@ export default class FluentUIReactIconsFontSubsettingPlugin implements BundlerPl
             }
           }
           const optimizationPromises: Promise<void>[] = [];
+          const packageToFontAssets = new Map<
+            string,
+            { usedExports: Set<string>; fontAssets: { assetName: string; codepoints: Record<string, number> }[] }
+          >();
 
           for (const pkgLibPath of unresolvableNamespacePackages) {
             // Sibling modules would otherwise subset this package's shared fonts down to *their*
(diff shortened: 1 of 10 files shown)


packages/react-icons-font-subsetting-webpack-plugin/src/index.ts
```diff
@@ -110,19 +131,49 @@ export default class FluentUIReactIconsFontSubsettingPlugin implements BundlerPl
...
-                    `to an emitted asset. Fonts will NOT be subset. Ensure a \`type: 'asset'\` (or 'asset/resource') ` +
-                    `module rule matches /\\.(ttf|woff2?)$/.`,
-                ),
-              );
-              continue;
+          const owningPackages = [...packageToFontAssets].filter(([, { fontAssets }]) => fontAssets.length > 0);
+          const orphanedPackages = [...packageToFontAssets].filter(([, { fontAssets }]) => fontAssets.length === 0);
+
+          if (owningPackages.length > 0 && orphanedPackages.length > 0) {
+            // Identical fonts across copies hash to one asset, which can name only one copy as its
```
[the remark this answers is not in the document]
Hotell: Correct, and reproduced before fixing. Fixed in ea4e3bf.

The scenario ran exactly as you described — named-import copy owns the shared asset, namespace copy is invisible to detection:

```
updatedAssets: [{"name":"Regular-0.ttf","size":660}]
```

660 bytes is the asset subset down to the owning copy's icons, with the namespace copy's glyphs deleted — while its own warning claimed the fonts had been left whole.

The root cause is that one set was answering two different questions. `packageToUsedFontExports` means *"packages eligible for subsetting"*, which is why namespace copies are deleted from it. Detection needs a different question: *"packages whose glyphs must survive"*, and that set includes them precisely because their glyphs cannot be enumerated.

So the two are now separate:

- **participants** (drives collision detection) = used-exports map keys ∪ `unresolvableNamespacePackages`
- **subsettable** (drives the subsetting loop) = used-exports map, unchanged

A namespace copy can therefore be *owning* for detection purposes while never being subset.

One deliberate detail beyond the report: the missing-asset-rule branch still keys on **subsettable only**. Had participants fed that branch too, a namespace-only copy whose fonts were not emitted would begin emitting a misleading asset-rule warning where it is currently silent.

This also fixed a second variant I found while tracing it — when the *namespace* copy owns the asset and the named copy is orphaned, the old code fell into the all-zero branch and blamed asset-rule configuration. Output was safe, diagnosis was wrong. Now reported as a collision.

Both are covered by tests that fail against the previous commit:

```
× counts a copy reached only through an unresolvable namespace import
  → expected [ Array(1) ] to deeply equal []
× reports a collision, not a missing asset rule, when the unresolvable copy owns the asset
  → expected 'FluentUIReactIconsFontSubsettingPlugi…' to contain 'installed more than once'
```

The existing single-package namespace test guards the other direction: one package with both import kinds gives `owning=1, orphaned=0`, so no collision warning appears.

packages/react-icons-font-subsetting-webpack-plugin/src/index.ts
```diff
@@ -108,21 +132,65 @@ export default class FluentUIReactIconsFontSubsettingPlugin implements BundlerPl
...
+
+          for (const pkgLibPath of participants) {
+            packageToFontAssets.set(
+              pkgLibPath,
+              await getFontAssetsAndCodepoints(pkgLibPath, compilation, compiler.context),
+            );
+          }
+
+          const owningPackages = [...packageToFontAssets].filter(([, fontAssets]) => fontAssets.length > 0);
+          const orphanedPackages = [...packageToFontAssets].filter(([, fontAssets]) => fontAssets.length === 0);
```
PaulGMardling: Could this package-level classification miss partial sharing between versions?

For example, if two versions share one byte-identical font asset but each also owns other distinct font assets, both would be included in `owningPackages` and neither in `orphanedPackages`.

Could the shared font then be subset using exports from only the package named by its `sourceFilename`, potentially dropping glyphs used through the other version?

Would this need collision detection per font asset instead?

Hotell: Good catch. Yes, owner/orphan classification misses distributed ownership when every copy owns an asset.

We now enforce the plugin’s actual invariant directly:

- One participating @fluentui/react-icons root: subset normally.
- Multiple roots: leave fonts un-subset and warn, or fail with onDuplicateInstances: 'error'.

This removes the need for per-asset collision detection.

https://github.com/microsoft/fluentui-system-icons/pull/1234/changes/ab5aadc0665c39809ffde78bce0aa4ffad6d7933

packages/react-icons-font-subsetting-webpack-plugin/src/index.ts
```diff
@@ -108,6 +130,29 @@ export default class FluentUIReactIconsFontSubsettingPlugin implements BundlerPl
             );
           }
 
+          // Namespace copies contribute no *known* glyphs, so they never reach the used-exports
+          // map — but their fonts still have to survive, which makes them part of this analysis.
+          const participants = new Set([...packageToUsedFontExports.keys(), ...unresolvableNamespacePackages]);
```
PaulGMardling: Is my understanding correct on the usage here?

Is `...packageToUsedFontExports.keys()` pulling out all the package paths recorded in the map, regardless of whether their export sets are empty? Using `'bar'` and `'some-module'` as simplified names…

is there a possible case where we may see something like

```import 'some-module';```

included as a key, where we're actually looking for packages that contribute used exports, such as `'foo'` from `'bar'`?

``` import { foo } from 'bar';```

So the plugin collects information like

```
packageToUsedFontExports = new Map([
  ['bar', new Set(['foo'])],
  ['some-module', new Set()],
]);
```

Here, `'bar'` contributes the used export `'foo'`. `'some-module'` contributes no used exports but is still present in the map.

```
[...packageToUsedFontExports.keys()] 
// -> ['bar', 'some-module']
```

Would this count `'some-module'` as a participant despite its empty export set, potentially triggering the duplicate warning or build failure?

Do we want to count two that go into the restaurant, one orders, the other doesn't, yet we're still saying they're both diners?

Comment (PaulGMardling) [reply packages/react-icons-font-subsetting-webpack-plugin/src/index.ts]: A related question: could one installation also count as two participants if application code uses ESM imports while a dependency uses CommonJS `require()`?

```
['/same-install/react-icons/lib', new Set(['GamesFilled'])]
['/same-install/react-icons/lib-cjs', new Set(['AddFilled'])]

```

Unlike the empty-set case above, both entries contribute icons, but they belong to the same installation. Could this also trigger the duplicate warning or build failure? Should their usage be combined under the common package root?

PaulGMardling (approved):

packages/react-icons-font-subsetting-webpack-plugin/src/index.ts
Hotell: Yes, lib and lib-cjs were being treated as separate participants even though they belong to the same installation. 

Mixed ESM/CommonJS usage can occur in practice, although consumers should avoid it where possible because CommonJS may reduce tree-shaking effectiveness. 

We now key usage by the package installation root, union exports from both module formats, and retain the individual output roots only for asset discovery. A real webpack/rspack fixture confirms ESM import and CommonJS require() can coexist without triggering duplicate detection.

On the empty-set question: a font module from a genuinely different installation still counts intentionally, because subsetting supports exactly one resolved package installation.

Fixed in 421220fdbda833220ca38c3bb0dfe1d7afcef70c

PaulGMardling: Thanks, everyone who enters the restaurant counts as a diner, even if they don’t order. Just don’t count the same person twice because they used both doors 😄 Got it!

PR #1242, opened 2026-09-11
fix(react-icons): export package metadata

Hotell:
## Summary

- expose `@fluentui/react-icons/package.json` through the package export map
- add regression coverage for the package metadata subpath

## Validation

- `corepack yarn nx run react-icons:build-verify -- --testNamePattern='should export package.json'`
- `corepack yarn nx run-many -t lint check:types-packaging -p react-icons`


PaulGMardling (approved):
Looks good! 
Nice, focused fix to expose the package metadata, with regression coverage and the ESM verifier updated to handle JSON appropriately. 👍

PR #1246, opened 2026-09-15
feat(react-icons-atomic-webpack-loader): add export-level atomization

Hotell:
## Summary

- add opt-in `moduleGranularity: 'icon'` while preserving the existing `'family'` default
- rewrite selected exports to versioned query-addressed ESM modules without publishing one physical file per icon export
- support static imports/re-exports, direct atomic imports, and statically provable dynamic-import groups
- add the complete `__fluentIcon=v1` compilation handshake required by font and SVG-sprite subsetting plugins
- preserve and compose source maps only when the bundler requests them
- cover Webpack and Rspack integration, system and brand icons, headless/font variants, exact queried identities, malformed selectors, and stale protocol contracts
- document loader-rule coverage, compatibility, rollback, caching, and chunk-distribution considerations

## Why icon granularity?

Family mode already tree-shakes unused exports. Icon mode changes module identity so exports from the same physical icon family can be placed in different chunks.

For example:

```js
// main.js
import { AddRegular } from '@fluentui/react-icons';

// lazy.js
import { AddFilled } from '@fluentui/react-icons';
```

With the default family granularity, both imports resolve to the same generated family module:

```text
main.js  ─┐
          ├─> lib/atoms/svg/add.js
lazy.js  ─┘
```

The bundler must assign that single module identity as a unit, which can move the family into a shared or otherwise less optimal chunk.

With `moduleGranularity: 'icon'`, the loader emits independent virtual module identities:

```text
main.js  ──> lib/atoms/svg/add.js?__fluentIcon=v1&export=416464526567756c6172
lazy.js  ──> lib/atoms/svg/add.js?__fluentIcon=v1&export=41646446696c6c6564
```

Each selected module contains only its declaration and required imports, allowing the bundler to keep `AddRegular` with the main chunk and `AddFilled` with the lazy chunk.

## Configuration

```js
{
  test: /\.[mc]?[jt]sx?$/,
  enforce: 'pre',
  use: [
    {
      loader: '@fluentui/react-icons-atomic-webpack-loader',
      options: {
        moduleGranularity: 'icon',
      },
    },
  ],
}
```

The rule must cover importers and the generated Fluent ESM atom directories because icon mode uses two loader passes:

1. Rewrite the importer to a query-addressed family request.
2. Process the resolved family file and emit the selected virtual module.

Font and SVG-sprite variants require their matching query-aware subsetting plugin. The compilation handshake fails closed when the plugin is missing or supports a different selector key/version. Family mode does not use this handshake.

## Validation

- builds, lint, unit tests, and Webpack/Rspack fixtures for the atomic loader, font plugin, and SVG-sprite plugin
- default family-mode and explicit icon-mode plugin integration coverage
- generated ESM declaration independence verification
- repeatable benchmark covering 5,000 selected exports and 20 complete importer rewrites

## Notes

- `moduleGranularity: 'family'` remains the default and rollback path.
- CommonJS selected-module emission is intentionally unsupported.
- Icon mode improves chunk distribution; it does not improve export tree-shaking over family mode.
- The general SVG-sprite per-entrypoint attribution fix was separated into #1247.


packages/react-icons-atomic-webpack-loader/test/selector-protocol.test.ts
```diff
@@ -0,0 +1,94 @@
...
+  it('preserves directives and imports while emitting exactly one declaration', () => {
+    const source = [
+      '"use client";',
+      "import { createFluentIcon } from '../../utils/createFluentIcon.js';",
+      "export const AddFilled = createFluentIcon('AddFilled', '1em', ['filled']);",
+      "export const AddRegular = createFluentIcon('AddRegular', '1em', ['regular']);",
+    ].join('\n');
+
+    const result = selectExports(source, resourcePath, { kind: 'export', exportName: 'AddFilled' });
+    expect(result.code).toContain('"use client"');
```
Hotell: use inline snapshot

packages/react-icons-atomic-webpack-loader/test/transform.test.ts
```diff
@@ -6,6 +6,80 @@ const transform = (source: string, iconVariant: IconVariant = 'svg', fallbackVar
...
+        [
+          `import { AddFilled } from '@fluentui/react-icons/svg/add?__fluentIcon=v1&export=41646446696c6c6564';`,
+          `import { bundleIcon } from '@fluentui/react-icons/utils';`,
+          `import { useIconContext } from '@fluentui/react-icons/providers';`,
+        ].join('\n'),
+      );
+    });
+
+    it('preserves explicit direct-path strategy while selecting the export', () => {
+      const { code } = transformIcons(`import { AddFilled } from '@fluentui/react-icons/svg/add';`, {
```
Hotell, line 30: lets also add test that uses iconVariants fonts and headless and assert against proper transform 

eg: `import { AddFilled } from '@fluentui/react-icons/headless/fonts/add?__fluentIcon=v1&export=41646446696c6c6564`

packages/react-icons-atomic-webpack-loader/src/selector-protocol.ts
```diff
@@ -0,0 +1,174 @@
...
+export function createExportSelector(exportName: string): string {
+  return `?__fluentIcon=${SELECTOR_PROTOCOL_VERSION}&export=${encodeExportName(exportName)}`;
+}
+
+export function createGroupSelector(exportNames: Iterable<string>): string {
+  const encoded = Array.from(new Set(Array.from(exportNames, encodeExportName))).sort();
+  if (encoded.length === 0) {
+    throw new Error('a selector group must contain at least one export');
+  }
+  return `?__fluentIcon=${SELECTOR_PROTOCOL_VERSION}&group=${encoded.join('.')}`;
```
Hotell: why is `__fluentIcon` not a const as well ?

packages/react-icons-atomic-webpack-loader/test/selector-protocol.test.ts
Hotell: we should test also against react-brand-icons

packages/react-icons-atomic-webpack-loader/test/make-configs.js
```diff
@@ -199,6 +202,23 @@ const entries = {
...
+    mustInclude: ['AddFilled', 'AddRegular', 'DrawImage24Filled'],
+    mustExclude: ['Add12Regular', 'DrawImageRegular'],
+  },
+  'icon-granularity-dynamic': {
+    src: './src/icon-granularity-dynamic.js',
+    loaderOptions: { moduleGranularity: 'icon', allowDynamicImports: true },
+    bundleIcons: true,
+    selectedModuleCount: 2,
+    selectedGroupCount: 1,
+    mustInclude: ['AddFilled', 'AddRegular'],
```
Hotell, line 220: can we test that there is the resource query or the bundle will already contain the virtualized per icon chunks ?

packages/react-icons-atomic-webpack-loader/CHANGELOG.md
```diff
@@ -1,3 +1,14 @@
+## Unreleased
```
Hotell: all changelogs should be reverted, those are created during release

packages/react-icons/build-verify.test.js
```diff
@@ -99,6 +99,48 @@ async function assertAtomOptionsArgument(atomDir) {
 }
 
 describe('Build Verification', () => {
+  describe('ESM atom selector contract', () => {
```
Hotell: do we need this ? we already have similar test in place

packages/react-icons-font-subsetting-webpack-plugin/test/make-configs.js
```diff
@@ -107,12 +107,11 @@ function createConfig(name, entry, adapter, isDevServer) {
                 // Rewrite barrel `@fluentui/react-icons` imports to headless font atoms
                 // before the bundler parses them.
                 test: /\.js$/,
-                include: resolve(__dirname, 'src'),
```
Hotell, line 110: why was this removed ?

packages/react-icons-atomic-webpack-loader/README.md
```diff
@@ -80,6 +80,91 @@ module.exports = {
...
+`@fluentui/react-brand-icons`. It does not need to process all of
+`node_modules`; an application that only rewrites its own source can use a
+targeted include:
+
+```js
+const path = require('path');
+
+{
+  test: /\.[mc]?[jt]sx?$/,
+  include: [
```
Hotell, line 116: this might not be enough as we need to transforms all icons including those imports that are already coming from 3rd packages tha live in node_modules

packages/react-icons-atomic-webpack-loader/test/selector-protocol.test.ts
Hotell: Added explicit `react-brand-icons` resource validation and selected-output snapshot coverage using `ProjectColor`.

packages/react-icons-atomic-webpack-loader/test/make-configs.js
Hotell: Strengthened this to assert the exact canonical export/group resource queries found in the Webpack and Rspack compilation module graphs. The emitted bundle does not retain resource-query text, so module identities are the direct assertion surface.

packages/react-icons-atomic-webpack-loader/README.md
Hotell: Updated the guidance to distinguish both modes: omit `include` to rewrite arbitrary third-party importers, or use a targeted list containing app source, every known dependency importer, and the mandatory Fluent atom directories for lower rule-matching overhead.

packages/react-icons/build-verify.test.js
Hotell: The existing tests sample generated files or verify export-to-family resolution, but they do not guard the selector requirement that every emitted ESM declaration be independent of sibling bindings. I retained that distinct all-files invariant and moved it into the existing `Atoms` suite so it is no longer a separate top-level suite.

packages/react-icons-font-subsetting-webpack-plugin/test/make-configs.js
Hotell: It was removed so the loader could run its required second pass over generated atom files. I restored `include` with the correct targeted scope: fixture source plus the Fluent generated `lib/atoms` directories.

packages/react-icons-atomic-webpack-loader/CHANGELOG.md
Hotell: Reverted all changelog changes in the loader and both subsetting plugins; they now match `upstream/main`.

packages/react-icons-atomic-webpack-loader/test/transform.test.ts
Hotell: Added an exact icon-granularity transform test for `iconVariant: "fonts"` with `headless: true`; it asserts the `headless/fonts/add` selector request.

packages/react-icons-atomic-webpack-loader/test/selector-protocol.test.ts
Hotell: Updated the selected ESM output assertions to use inline snapshots.

packages/react-icons-atomic-webpack-loader/src/selector-protocol.ts
Hotell: Added `SELECTOR_QUERY_KEY` and reused it for selector generation, detection, validation, and protocol lookup.

Hotell:
Split the general SVG-sprite per-entrypoint attribution fix into #1247. This PR now remains focused on icon granularity and the selector protocol handshake.

packages/react-icons-atomic-webpack-loader/src/select-export.ts
```diff
@@ -0,0 +1,214 @@
...
+  generateSourceMap = true,
+): { code: string; map: ReturnType<MagicString['generateMap']> | undefined } {
+  if (collectMetrics) {
+    metrics.emissions++;
+  }
+  if (selector.kind === 'group') {
+    const filename = basename(resourcePath);
+    const code = selector.exportNames
+      .map((exportName) => `export { ${exportName} } from './${filename}${createExportSelector(exportName)}';`)
+      .join('\n');
```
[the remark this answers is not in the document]
Hotell: The selected per-icon target modules retain their own "use client" directives. A server-evaluated re-export wrapper exposes those imports as client references without needing to become a client boundary itself, so the group remains pure re-exports. The group-emission test now snapshots that exact output.

packages/react-icons-atomic-webpack-loader/src/index.ts
```diff
@@ -62,18 +71,55 @@ export interface FluentIconsAtomicImportLoaderOptions {
...
+    }
+  } catch (error) {
+    const reason = error instanceof Error ? error.message : String(error);
+    return this.callback(
+      new Error(`FluentIconsAtomicImportLoader: Failed to select "${resourcePath}${resourceQuery}": ${reason}`),
+    );
+  }
+
+  if (isGeneratedIconPackageResource(resourcePath)) {
+    return this.callback(null, sourceCode, inputSourceMap);
```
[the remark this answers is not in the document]
Hotell: Fixed in 8c7d72380d. Both generated-package and source-text no-op exits now return the incoming map only when source maps are enabled, with regression coverage for both paths.

packages/react-icons-atomic-webpack-loader/src/select-export.ts
dmytrokirpa, line 49: Group selectors are used by the statically provable dynamic-import path, but this branch emits only bare re-exports and drops the generated family's leading "use client" directive. A grouped module can therefore cross an RSC boundary as a server module even though the selected icon declarations are client modules. Please preserve the directive in the group-emission output and add/keep a regression test covering the exact grouped output.

packages/react-icons-atomic-webpack-loader/src/selector-protocol.ts
```diff
@@ -0,0 +1,177 @@
...
+
+export const SELECTOR_PROTOCOL_VERSION = 'v1';
+export const SELECTOR_CACHE_SALT = `fluent-icon-selector-${SELECTOR_PROTOCOL_VERSION}`;
+export const SELECTOR_QUERY_KEY = '__fluentIcon';
+export const SELECTOR_PROTOCOL_IDENTIFIER = `${SELECTOR_QUERY_KEY}=${SELECTOR_PROTOCOL_VERSION}`;
+export const SELECTOR_CAPABILITY = Symbol.for(`fluentui.react-icons.selector-protocol:${SELECTOR_PROTOCOL_IDENTIFIER}`);
+
+const IDENTIFIER_PATTERN = /^[A-Za-z_$][A-Za-z0-9_$]*$/;
+const HEX_PATTERN = /^(?:[0-9a-f]{2})+$/;
+const SELECTOR_QUERY_PATTERN = new RegExp(`(?:^\\?|&)${SELECTOR_QUERY_KEY}=`);
```
dmytrokirpa, line 11: The detector only recognizes `__fluentIcon=`. A malformed reserved query such as `?__fluentIcon` (or `?foo=1&__fluentIcon`) returns null and then falls through to the generated-resource pass-through, silently loading the full family instead of failing closed. Please detect the reserved key even without `=` so malformed selector requests are rejected as documented.

packages/react-icons-atomic-webpack-loader/src/transform.ts
```diff
@@ -385,7 +510,7 @@ export function transformSource(source: string, options: TransformOptions): Tran
 
   return {
     code: src.toString(),
-    map: src.generateMap({ hires: true }),
+    map: sourceMap ? src.generateMap({ hires: true }) : undefined,
```
dmytrokirpa, line 513: The ordinary importer rewrite map is generated without a source path or embedded source content, while the selector-emission map supplies both. When this map is composed with an incoming source map, importer positions may be unmapped or associated with an empty source. Please generate this map with the current path and source content (consistent with the selector-emission path).

dmytrokirpa (approved):

packages/react-icons-atomic-webpack-loader/src/select-export.ts
Hotell: This is the same concern as the earlier [resolved thread](https://github.com/microsoft/fluentui-system-icons/pull/1246#discussion_r4038705056). The group module intentionally contains only re-exports; each `./add.js?__fluentIcon=v1&export=...` request is processed separately, and its selected icon module retains the generated family's leading `"use client"` directive. A server-evaluated re-export wrapper therefore exposes client references without itself becoming a client boundary. I also updated the group-emission inline snapshot test to use client-marked family source, making the emitted wrapper (and absence of a directive there) explicit. This verifies the emitted modules, though we do not currently have a Next.js RSC integration test for this path.

PR #1248, opened 2026-09-18
feat: add Fabric Ontology file type icon

bigbadcapers:
Adds FileIconType.fabricOntology and its resolver mapping for the Fabric Ontology SVG assets from fabric-cdn 4.0.7. Updates the default CDN URL to fabric-cdn-prod_20260917.001, assuming fabric-cdn run 56984169 completes successfully. Validation: react-icons-file-type tests (42), build, and lint passed.


Hotell (approved):
LGTM

